PR #43, opened 2022-04-13
Close #42

### LINKED ISSUE #42: `typescript-transform-paths` to be optional as it's too slow
I've adopted the `typescript-transform-paths` library to support the path alias - #20. However, when building an SDK libary with the `typescript-transform-paths`, elapsed time be about three times.

Therefore, it would better to turn it on only when the alias path has been configured. In the same reason, `typescript-is` would be turned on only when the assert mode in on - #32.

samchon:
Ready to publish the v2.0.6 module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,11 @@ The `npx nestia install` command installs those dependencies with `package.json`
 export interface IConfiguration
 {
     /**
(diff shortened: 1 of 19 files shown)


samchon (commented):
OK

PR #69, opened 2022-05-03
V2.1, support `swagger.json` build

samchon:
From now on, you can generate the `swagger.json`, too.

Of course, it does not require any extra swagger comment and DTO decorator, either.


samchon (commented):
OK, prepared to publish the v2.1 update

PR #88, opened 2022-05-15
Fix: broken bash command in `README.md`

BWsix:
Thank you for creating this awesome package!
I ran the command copied from the README.md and found that the `--out` tag is broken, so I updated the README.md.
(Sorry for letting the formatter modify the last line of this file, while I was editing directly on Github)


samchon (approved):
Thanks for emension

PR #89, opened 2022-05-17
Close #87, revise README

### LINKED ISSUE #87: Revise README and WIKI once more
https://github.com/nestjs/docs.nestjs.com/pull/2304

Reviews from the `docs.nestjs.com` is impressive and there are a lot of things to learn. Revise this README and WIKI content following the advises. Especially, make something like "Getting Started" content in the top of the document for such reason:

> The block of scripts that has nestia generating files is cool, but I'd like there to be explanation as to what they do rather than just "run these scripts". Too many people copy-paste-run without reading, so if the large block is broken up and explained then 1) it's harder to run it all together 2) we can have explanations for what each script does more easily which 3) let's us say "did you read what that script does?" when someone complains that the script trashed their machine (not that it's going to, but it could have unintended side effects of some sort that the dev was not aware of even though we outlined what's happening).

samchon:
To be concise


samchon (commented):
OK

PR #100, opened 2022-05-25
Fix #99

### LINKED ISSUE #99: Swagger generates a wrong path parameter
When define a controller method parameter which has different name with the binded argument, `nestia` follows the parameter name and it causes error on the Swagger Editor. For example, if define a `value` named parameter with `flag` path binding, `nestia` generates not `flag` argument, but `value` argument.

```typescript
export class SomeController
{
    @helper.TypedRoute.Get(":flag")
    public someMethod(@helper.TypedParam("flag", "boolean") value: boolean): ISomeResponse;
}
```

samchon:
Endless swagger generator editions

### DIFF
--- a/demo/array/swagger.json
+++ b/demo/array/swagger.json
@@ -203,7 +203,6 @@
   "components": {
     "schemas": {
       "ISaleArticleComment.IStore": {
-        "$id": "ISaleArticleComment.IStore",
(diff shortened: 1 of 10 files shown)


samchon (commented):
OK, finally completed

PR #107, opened 2022-05-31
rename TypeFactory

MJoon-Jung:
Fix fiile name
src/factories/TypeFactry.ts -> src/factories/TypeFactory.ts


samchon (changes requested):
Thanks for changing.

The namespace `TypeFactory` is not being used by no one.

Can you commit this PR again by erasing the entire `factories` directory?

PR #120, opened 2022-07-02
Fix #117

### LINKED ISSUE #117: Don't use TypeScript 4.7 and rollback to 4.6
`TypeFactory` type `typescript-json` breaks on TS 4.7 version.

TS 4.7 can't get properties of interface and failed to get parent module of each interface.

Therefore, roll back to 4.7 and look forward the bug to be fixed.

samchon:
Fix dependencies' versions for special error case #117

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "nestia",
-  "version": "3.0.4",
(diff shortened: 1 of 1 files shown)


samchon (commented):
OK

PR #130, opened 2022-07-19
Fix #129 - Correct typing for compilerOptions

### LINKED ISSUE #129: IConfiguration should not use ts.CompilerOptions
### Summary
  - **SDK Version**: 3.0.8
  - **Expected behavior**: accept `compilerOptions` as the same interface used in `tsconfig.json`
  - **Actual behavior**: the type `TsConfig`, which is internal usage only, is not compatible with `CompilerOptions`

### Code occuring the bug
```typescript
import type nestia from 'nestia';

export const NESTIA_CONFIG: nestia.IConfiguration = {
  input: [
    './src/app/*.controller.ts',
    './src/modules/**/*.controller.ts',
  ],
  output: './src/sdk',
  json: true,
  swagger: {  output: './src/assets/api-docs.json' },
  compilerOptions: {
    module: 'commonjs',
  },
};

export default NESTIA_CONFIG;
```

_error_
```
error TS5024: Compiler option 'module' requires a value of type string.
```

SkyaTura:
Fix #129

### DIFF
--- a/src/IConfiguration.ts
+++ b/src/IConfiguration.ts
@@ -1,4 +1,5 @@
 import ts from "typescript";
+import type { StripEnums } from "./utils/StripEnums";
 
 /**
  * Definition for the `nestia.config.ts` file.
@@ -38,7 +39,7 @@ export interface IConfiguration {
      * }
      * ```
      */
-    compilerOptions?: ts.CompilerOptions;
+    compilerOptions?: StripEnums<ts.CompilerOptions>;
 
     /**
      * Whether to assert parameter types or not.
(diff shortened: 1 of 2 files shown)


samchon:
Will merge after test automation program implementation.

Thanks for your contribution.

samchon (commented):
May I ask you the reason why wrapping tsconfig object?

SkyaTura:
> May I ask you the reason why wrapping tsconfig object?

When using TypeScript programmatically, the config must be similar to the structure used in tsconfig.json, however, that type isn't exposed in the typescript package, and cannot be extended.

This being said, if you use the `CompilerOptions` available, trying to change properties that uses enums will raise a type error when compiling nestia.config.ts. However, if you use the exposed enums instead, you get an error when instantiating the compiler, which is expecting a string, not a number.

---

Sorry for the late response and not writing those tests, these last days were rushed for me. I'll be more cautious next time.

I loved this project, and, if you don't mind, I intend to submit more features and tweaks.

PR #145, opened 2022-08-20
Close #144, `IConfiguration.primitive`

### LINKED ISSUE #144: Add new configuration property, `primitive`
Until now, `nestia` generated SDK libraries had capsuled DTOs through `Primitive` type. 

For reference, the `Primitive` type has designed to convert class type to primitive interface type, by erasing methods. Also, `Primitive` converts capsuled DTO type to its returned type of `toJSON()` method when the `toJSON()` method exists. `nestia` generated SDk libraries had utilized it because Rest API requires only the primitve object value.

However, I've listened that some backend developers design custom DTO classes to make client developers to be convenient. In that case, the atomatic `Primitive` type wrapping makes them inconvenient. Therefore, from now on, I'll provide a new configuration property, `primitive` and allow user of `nestia` to turn off the automatic `Primitive` type wrapping.

```typescript
export function store
    (
        connection: IConnection,
        section: string,
        saleId: string,
        input: Primitive<store.Input>
    ): Promise<store.Output>
{
    return Fetcher.fetch
    (
        connection,
        store.ENCRYPTED,
        store.METHOD,
        store.path(section, saleId),
        input
    );
}
export namespace store
{
    export type Input = Primitive<ISaleReview.IStore>;
    export type Output = Primitive<ISaleInquiry<ISaleReview.IContent>>;

    export const METHOD = "POST" as const;
    export const PATH: string = "/consumers/:section/sales/:saleId/reviews";
        request: false,
        response: false,
    };

    export function path(section: string, saleId: string): string
    {
        return `/consumers/${encodeURIComponent(section)}/sales/${encodeURIComponent(saleId)}/reviews`;
    }
}
```

samchon:


samchon (commented):
OK, no problem.

PR #146, opened 2022-08-20
Close #143, support regex path

### LINKED ISSUE #143: Support regex path
### Summary
Validating path arguments does not take into account possible regex in the route parameters.

For example, the following endpoint:
```
  @Get(':id([0-9a-f]{24})')
  async getCall(
    @Req() req: Request,
    @Param('id') id: string,
  ): Promise<CallDTO> {
```
Will fail when generating the swagger json spec with `Error on CallController.getCall(): binded arguments in the "path" between function's decorator and parameters' decorators are different (function: [id([0-9a-f]{24})], parameters: [id])`

samchon:


samchon (commented):
OK, no problem.

Issue #216, opened 2023-01-16
Support generating swagger format YAML


sacru2red:
## Feature Request
Thanks for your suggestion. Feel free to just state your idea. Writing the issue, it would better to filling the below items:

  - current @nestia/sdk generates swagger only json
  - generally, use 'swagger-ui-express' lib or others for serve swagger document,
I had try to serve exported swagger.json from @nestia/sdk but can't
then, I try yaml file. It's Ok
So, Support generating swagger format YAML

sacru2red:
example

src/main.ts
```
import 'reflect-metadata'
import { NestFactory } from '@nestjs/core'
import { AppModule } from './app.module'
import { NestExpressApplication } from '@nestjs/platform-express'
import * as swaggerUi from 'swagger-ui-express'
import * as path from 'path'
import * as Express from 'express'
import * as YAML from 'yamljs'

async function bootstrap () {
  const app = await NestFactory.create<NestExpressApplication>(AppModule)

  const swaggerRouter = Express.Router()
  const swaggerOpts = {
    explorer: true,
  }
  try {
    const swaggerYaml = YAML.load(path.join(__dirname, '..', '..', 'swagger', 'swagger.yaml'))
    swaggerRouter.use('/api-docs', swaggerUi.serve)
    swaggerRouter.get('/api-docs', swaggerUi.setup(swaggerYaml, swaggerOpts))
    app.use(swaggerRouter)
  } catch (error) {
    if (error instanceof Error) {
      console.log('fail to register swagger router check swagger aseet is builded', error.stack)
    }
  }

  const port = Number(process.env.PORT ?? 5000)
  await app.listen(port)
}
void bootstrap()
```

Aderinom:
Heyo sacru2red! 
I'm not a maintainer - but saw your issue here.

I personally just use the @nestjs/swagger module as follows:
```
import * as Doc from './swagger.json';

export function addSwagger(app: INestApplication) {
  SwaggerModule.setup('api', app, Doc, {
    customCss: swaggerDarkCss,
  });
}
```
Works perfectly and I can get the Yaml over the /api-yaml endpoint f.e.

Alternatively you could also write a small script for the swagger generation as follows:
```
import { execSync } from 'node:child_process';
import { writeFileSync } from 'node:fs';

execSync(`npx nestia swagger`, {
  stdio: 'inherit',
});

import * as doc from '../dist/swagger.json';

const doced = structuredClone(doc); // make an editable clone
doced.servers = [];

[...]

PR #217, opened 2023-01-16
support to generate swagger YAML format

sacru2red:
#216


samchon:
Sorry for late response. 

No plan to support YAML.

What about converting to the YAML by yourself through batch script?

PR #247, opened 2023-02-11
fix(@nestia/fetcher): incorrect overloading of fetcher functions

### LINKED ISSUE #246: Incorrect Overloading of Fetcher Functions
### Summary
Write a short summary of the bug in here.

  - **SDK Version**: main
  - **Expected behavior**: correct Overloading of Fetcher Functions
  - **Actual behavior**: Incorrect Overloading of Fetcher Functions

https://github.com/samchon/nestia/blob/45be5a1937b20244d347b740eeaa5af73a84c987/packages/fetcher/src/Fetcher.ts#L35-L40

In https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE, `Request body` is allowed, but in `@nestia/fetcher`, it becomes the same as `GET`, which disallows the `Request body`

So I think, `DELETE` method should be added in https://github.com/samchon/nestia/blob/45be5a1937b20244d347b740eeaa5af73a84c987/packages/fetcher/src/Fetcher.ts#L56

wibus-wee:
fix #246

### DIFF
--- a/packages/fetcher/src/Fetcher.ts
+++ b/packages/fetcher/src/Fetcher.ts
@@ -24,7 +24,7 @@ import { Singleton } from "./internal/Singleton";
  */
 export class Fetcher {
     /**
-     * Fetch function for the `GET` or `DELETE` methods.
+     * Fetch function for the `GET` methods.
      *
      * @param connection Connection information for the remote HTTP server
      * @param encrypted Whether the request/response body be encrypted or not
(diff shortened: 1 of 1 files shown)


samchon:
Will study about this issue at this Sunday. Sorry for waiting for a long time.

samchon (commented):
Will reflect your suggestion with a little bit editing.

Thanks for contribution,

PR #269, opened 2023-03-06
Fix #268 - when optional query param comes

### LINKED ISSUE #268: Optional query is assigned "undefined" string
findMany api has a optional query 'page'.
but if page is undefined, url is written as "/products?page=undefined". and this query value is string "undefined".

i think, "/products?" or "/products" is right.
(so i think, undefined type should be handled separately.)

In Korea,
 
findMany api를 구현했습니다. page는 optional한 쿼리입니다. sdk를 통해 findMany(connection)를 실행하면 undefined가 문자열 "undefined"로 전달되는 버그가 있습니다. ( ex) "/products?page=undefined" )
 이는 URLSearchParams의 처리 방식 때문인 것 같습니다. 제 생각에는 undefined값은 키조차 UrlSearchParams 객체 생성 인자로 전달되지 않아야 할 것 같습니다. ( ex) "/products?", "/products" )

```typescript
// This is nestia sdk.
export namespace findMany
{
    export type Output = PaginatedResponse<ProductSchema.General>;

    export const PATH: string = "/products";
        request: false,
        response: false,
    };

    export function path(page: number | undefined): string
    {
        return `/products?${new URLSearchParams(
        {
            page
        } as any).toString()}`;
    }
}
```

samchon:

```bash
# COMPILE THE BACKEND SERVER
npm run build

npm run test
```

### DIFF
--- a/packages/sdk/demo/union/src/api/functional/consumers/sales/entire_articles/index.ts
+++ b/packages/sdk/demo/union/src/api/functional/consumers/sales/entire_articles/index.ts
@@ -64,12 +64,13 @@ export namespace index
 
     export function path(section: string, saleId: number, ipAddr: string | undefined, href: string | undefined, query: ISaleReview.IQuery): string
     {
-        return `/consumers/${encodeURIComponent(section)}/sales/${encodeURIComponent(saleId)}/entire_articles?${new URLSearchParams(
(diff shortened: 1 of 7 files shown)


samchon (commented):
OK

Issue #282, opened 2023-03-16
Too long return type fails to generate SDK


kakasoo:
Hello, I brought a new issue today. :(

Now I have implemented a type called `Try<T,E>`. I defined a type that returns T in normal cases and E in other cases. I decided to treat errors as an object instead of throwing it so that even all errors could be inferred.
Therefore, this type of `Try<T,E>` is deduced to be a very long return type.

The problem happened here.
If you look at the code below, you can immediately see how Nestia generated the SDK.

```typescript
  @TypedRoute.Get('omg')
  async ThisFunctionHasTooLongReturnType(): Promise<
    Try<
      true,
      | typeof ERROR.TOO_LONG_KEY_NAME1
      | typeof ERROR.TOO_LONG_KEY_NAME2
      | typeof ERROR.TOO_LONG_KEY_NAME3
      | typeof ERROR.TOO_LONG_KEY_NAME4
      | typeof ERROR.TOO_LONG_KEY_NAME5
    >
  > {
    /**
     * If it's not zero, then return ERROR object.
     */
    let something: number = 0;
    if (something === 0) {
      return { result: true, code: 1000, data: true };
    } else if (something === 1) {
      return ERROR.TOO_LONG_KEY_NAME1;
    } else if (something === 2) {
      return ERROR.TOO_LONG_KEY_NAME2;
    } else if (something === 3) {
      return ERROR.TOO_LONG_KEY_NAME3;
    } else if (something === 4) {
      return ERROR.TOO_LONG_KEY_NAME4;
    } else if (something === 5) {
      return ERROR.TOO_LONG_KEY_NAME5;
    }
  }

```

```typescript
export namespace ThisFunctionHasTooLongReturnType
{
    export type Output = Primitive<Try<true, { readonly result: false; readonly code: 4000; readonly data: "Error happens something1."; } | { readonly result: false; readonly code: 4000; readonly data: "Error happens something2."; } | { ...; } | { ...; } | { ...; }>>;

    export const METHOD = "GET" as const;
    export const PATH: string = "/omg";
    export const ENCRYPTED: Fetcher.IEncrypted = {
        request: false,
        response: false,
    };

    export function path(): string
    {
        return `/omg`;
    }
}
```

Please check the Output type of the namespace "ThisFunctionHasTooLongReturnType".
At the end of the type, `"{...; } | {...; } | { ...; }".`

## if you check this...

```bash
git clone https://github.com/kakasoo/nestia-demo
cd nestia-demo
[...]

PR #284, opened 2023-03-19
fix : Too long return type fails to generate SDK #282

kakasoo:
- ts.TypeFormatFlags.NoTruncation apply
- I don't know how to test it work correctly, so I need your help.

I have previously said in issue #282 that "the problem is that too long types are not properly deduced". This was very important to me, so I analyzed the TypeScript compiler and solved the problem. This problem was too simple. The problem was solved by simply passing the flag to the TypeScript compiler. In my code this worked perfectly. Unfortunately, however, I don't know if this works only with my code or how I can test the nestia library.
So I have no choice but to ask for your help for the next time. I would appreciate it if you could check the PR and reflect it if there is no problem.

Thank you always for making such a wonderful library.

### DIFF
--- a/packages/sdk/src/analyses/ImportAnalyzer.ts
+++ b/packages/sdk/src/analyses/ImportAnalyzer.ts
@@ -73,7 +73,7 @@ export namespace ImportAnalyzer {
         const symbol: ts.Symbol | undefined =
             type.aliasSymbol || type.getSymbol();
         if (symbol === undefined)
-            return checker.typeToString(type, undefined, undefined);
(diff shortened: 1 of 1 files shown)


samchon (approved):
Good point.

samchon:
@kakasoo Upgrade to `@nestia/sdk@1.0.17`, then your problem would be resolved.

kakasoo:
> @kakasoo Upgrade to `@nestia/sdk@1.0.17`, then your problem would be resolved.

Thx. This was absolutely necessary for me. Good night!

PR #286, opened 2023-03-27
Fix #282 and Complement #284 - full name typing

### LINKED ISSUE #282: Too long return type fails to generate SDK

samchon:
Solution by @kakasoo


samchon (commented):
OK, working well.

PR #294, opened 2023-03-29
Fix #293 - type with aliasSymbol not defined

### LINKED ISSUE #293: invalid type(_object) is generated in sdk.
sdk doesn't generate Output type properly in specific case.

```typescript
interface Pagination{
    page: number;
}

interface Try<T> {
    code: 1000;
    data: T;
}
interface IException {
    code: number;
    data: string;
}

type TryCatch<T,E extends IException> = Try<T> | E;
const exception1 = {code:4040, data:"test message"} as const;
const exception2 = {code:4030, data:"test message2"} as const;

// I think this is bug!
TryCatch<Pagination, typeof exception1> // this type doesn't generated properly.
// result: in sdk file, type Output = TryCatch<Pagination, __object>



TryCatch<Pagination, typeof exception1 | typeof exception2> // this type generated properly.
// result: in sdk file, type Output = TryCatch<Pagination, {code:4040, data:"test message"} |  {code:4030, data:"test message2"} >
```

and it's really my code.
```typescript
// in contoller
  /**
   * 전체 상품 목록 조회
   *
   * @summary 상품 목록 조회 API
   * @tag products
   * @param page 페이지 정보 1이상의 정수, default 1
   * @returns 상품 목록
   * @throw 4002 유효하지 않은 query입니다.
   */
  @Get()
  async findMany(
    @Query('page') page?: string,
  ): Promise<
    TryCatch<
      PaginatedResponse<IProduct.Summary>,
      typeof Exception.INVALID_QUERY
    >
  > {
    const npage = Number(page ?? 1);
    if (isNaN(npage) || !typia.is<Page>({ page: npage }))
      return Exception.INVALID_QUERY;
    return ProductUsecase.findMany(npage);
  }

// in sdk file,
 export type Output = TryCatch<PaginatedResponse<IProduct.Summary>, __object>;

```

kakasoo:

```bash
# COMPILE THE BACKEND SERVER
npm run build

npm run test
```


packages/sdk/src/analyses/ImportAnalyzer.ts
```diff
@@ -79,21 +79,28 @@ export namespace ImportAnalyzer {
...
-                .map((child) =>
-                    explore_escaped_name(
-                        checker,
-                        genericDict,
-                        importDict,
-                        child,
-                    ),
-                )
-                .join(joiner);
+        else if (type.aliasSymbol === undefined) {
```
samchon, line 82: This logic would be fail when literal object expression requires an import statement:

```typescript
export const ERROR = {
    TOO_LONG_KEY_NAME1: {
        result: false,
        code: 4000,
        data: "Error happens something1.",
        detail: {
            code: 4000,
            message: "Error happens something1.",
        } as IError,
    },
];

@nest.Controller()
export class AppController {
    @core.TypedRoute.Get("singular")
    public async singular(): Promise<
        Try<true, typeof ERROR.TOO_LONG_KEY_NAME1>
    > {
        return { result: true, code: 1000, data: true };
    }
}
```
```

Comment (samchon) [reply packages/sdk/src/analyses/ImportAnalyzer.ts:82]: Generated SDK functional code missing `IError` import statement:

```typescript
import { Fetcher, Primitive } from "@nestia/fetcher";
import type { IConnection } from "@nestia/fetcher";

import type { Try } from "./../../controllers/AppController";

/**
 * @controller AppController.singular()
 * @path GET /singular
 * @nestia Generated by Nestia - https://github.com/samchon/nestia
 */
export function singular(connection: IConnection): Promise<singular.Output> {
    return Fetcher.fetch(
        connection,
        singular.ENCRYPTED,
        singular.METHOD,
        singular.path(),
    );
}
export namespace singular {
    export type Output = Primitive<
        Try<
            true,
            {
                readonly result: false;
                readonly code: 4000;
                readonly data: "Error happens something1.";
                readonly detail: IError;
            }
        >
    >;

    export const PATH: string = "/singular";
        request: false,
        response: false,
    };

    export function path(): string {
        return `/singular`;
    }
}
```

samchon (changes requested):
It would better to prohibit typeof statement, or find another way.

packages/sdk/src/analyses/ImportAnalyzer.ts
kakasoo: Thank you for providing the opposite case.

I'll think about another way. If import statements are a problem now, I will find out how to find that type of directory location.

Banning typeof keyword is simple, but I'm afraid it'll reduce Nestia's versatility. If developers have more to modify their own code to use Nestia, there will be more to give up on using it.
We need to minimize project code modification and make nestia available immediately.

It takes a lot of time.

PR #309, opened 2023-04-13
Add `peerDependencies` in core and sdk

samchon:
Succeeded to complicate dependency relationship of NestJS.

Therefore, restrict dependencies of NestJS through `peerDependencies`

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@nestia/core",
-  "version": "1.0.19",
+  "version": "1.0.20",
   "description": "Super-fast validation decorators of NestJS",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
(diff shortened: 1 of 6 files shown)


packages/core/src/decorators/EncryptedModule.ts
samchon, line 1: Unfortunately, older version of `@nestjs/common` had not exported `ModuleMetadata`.

Using `Parameters` TMP type is for such case.

packages/core/src/decorators/EncryptedRoute.ts
samchon, line 1: Older version of `rxjs` had not exported `Observable`, and had to import through `rxjs/operators` file.

packages/sdk/package.json
samchon, line 1: `ts-node` had worked well until `10.6.0` patch.

packages/core/package.json
samchon, line 1: This is result of analyses of dependency relationships of NestJS.

samchon (commented):
OK, no problem.

PR #310, opened 2023-04-13
Close #300 - support `@nest.HttpCode()` function

### LINKED ISSUE #300: How can i change http status code in swagger.
Post 메소드를 사용하지만 200 코드를 받고 싶은 경우처럼 기존과 다른 상태값을 받고 싶을 때, 어떻게 swagger 문서의 상태 코드를 변경할 수 있나요?

how can i change successful response status code in swagger?

```typescript
/**
* @return this is description for response
*/
@HttpCode(HttpStatus.OK)
@Post('test')
async test():Promise<number>{
  return 1;
}
```

If i use HttpCode decorator, status code will changed.
but swagger doesn't changed.

samchon:
When `@nest.HttpCode()` being used, it would be adjusted into both swagger and SDK

### DIFF
--- a/packages/sdk/demo/status/src/controllers/StatusController.ts
+++ b/packages/sdk/demo/status/src/controllers/StatusController.ts
@@ -1,9 +1,10 @@
+import core from "@nestia/core";
 import * as nest from "@nestjs/common";
 
 @nest.Controller()
 export class StatusController {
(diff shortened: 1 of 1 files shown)


samchon:
Bug of Github, I should send another PR for release note.

PR #312, opened 2023-04-14
Close #311 - enhance `TypedParam`'s type support

### LINKED ISSUE #311: Let `TypedParam` to check real parameter type
It would better to let `TypedParam` to check real parameter type and occurs compilation error when mismatched.

Also, if do not specify parameter type in decorator, just define it automatically through transformation.

```typescript
@Controller()
class SomeController {
    @Get("someMethod/:a/:b/:c")
    public someMethod(
        @TypedParam("a", "string") a: string, // no problem
        @TypedParam("b", "number") b: string, // occurs compilation error
        @TypedParam("c") c: number, // change to `@TypedParam("c", "number")` in compilation level
    ): void {}
}
```

samchon:
Changed `TypedParam` like below to enhance type support:

```typescript
/**
 * Type safe URL parameter decorator.
 *
 * `TypedParam` is a decorator function getting specific typed parameter from the
 * HTTP request URL. It's almost same with the {@link nest.Param}, but `TypedParam`
 * automatically casts parameter value to be following its type. Beside, the
 * {@link nest.Param} always parses all of the parameters as `string` type.
 *
 * Note that, if you just omit the `type` and `nullable` parameters, then
 * `TypedParam` automatically determines the `type` and `nullable` values
 * just by analyzing the parameter type. Only when you need to specify them
 * are, you want to use the `uuid` type.
 *
 * ```typescript
 * \@TypedRoute.Get("shopping/sales/:id/:no/:paused")
 * public async pause
 *     (
 *         \@TypedParam("id", "uuid"), id: string, // uuid specification
 *         \@TypedParam("no") id: number, // auto casting
 *         \@TypedParam("paused") paused: boolean | null // auto casting
 *     ): Promise<void>;
 * ```
 *
 * @param name URL Parameter name
 * @param type If omit, automatically determined by the parameter type.
 * @param nullable If omit, automatically determined by the parameter type.
 * @returns Parameter decorator
 *
 * @author Jeongho Nam - https://github.com/samchon
 */
export function TypedParam(
    name: string,
    type?: "boolean" | "number" | "string" | "uuid",
    nullable?: false | true,
): ParameterDecorator;
```


samchon (commented):
OK, working well.

PR #316, opened 2023-04-25
Setup CLI description

plrs9816:


samchon:
Thanks for detailed description, but I'm planning to abandon `ttypescript` and use only `ts-patch`.

It's because `ttypescript` has stopped maintaining, therefore cannot support TS@5 through `ttypescript`.

I'm sorry to cannot accept.

PR #342, opened 2023-05-13
Benchmark program: `nestia` vs `nestjs`

samchon:
Developed benchmark program comparing Nestia vs NestJS.

As `nestia` utilizes `typia`, and `nestjs` utilizes `class-validator`, benchmark result may almost same with `typia` vs `class-validator` (https://typia.io/docs/json/stringify/#server-performance). However, some users of `nestia` want to see the benchmark result not of `typia`, but of `nestia`. So, I developed and measured it directly.

As a result of such benchmark, I can say that `nestia` boosts up NestJS performance about 10 to 20x. About this benchmark result, I'll write an article and post it onto both https://nestia.io and https://dev.to

stringify%20Core(TM)%20i5-1135G7%20%40%202.40GHz/images/stringify.svg)


samchon:
It would better to erase `express-pure`.

PR #351, opened 2023-05-23
Add "date" type in `@TypedParam`

samchon:
From now on, you can use `"date"` on `@TypedParam` decorator.

It validates whether parameter value is following `YYYY-MM-DD` type or not.

```typescript
import core from "@nestia/core";

@Controller("param")
export class TypedParamController {
    @core.TypedRoute.Get(":value/date")
    public date(@core.TypedParam("value", "date") value: string): string {
        return value;
    }
}
```


kakasoo:
awesome!

Issue #411, opened 2023-06-20
[Suggestion] Add credentials propery to IConnection 


SeungjunWe:
## Feature Request
add [credentials property](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#sending_a_request_with_credentials_included) to IConnection
```typescript
export interface IConnection {
  credentials?: RequestCredentials;
}
```

## Examples of usage include
```typescript
import { functional } from 'sdk-created-by-nestia''

functional.user.getInfo({
  host: 'http://127.0.0.1',
  credentials: 'include',
})
```

samchon:
Published as v1.4.0 update.

PR #412, opened 2023-06-20
Add credentials to fetcher IConnection #411

SeungjunWe:
# Description
- Edit a suggestion #412

```bash
# COMPILE THE BACKEND SERVER
npm run build

npm run test
```

### DIFF
--- a/packages/fetcher/src/IConnection.ts
+++ b/packages/fetcher/src/IConnection.ts
@@ -47,4 +47,9 @@ export interface IConnection {
      * @default false
      */
     simulate?: boolean | Partial<IRandomGenerator>;
+
+    /**
+     * A string indicating whether credentials will be sent with the request always, never, or only when sent to a same-origin URL. Sets request's credentials.
+     */
+    credentials?: RequestCredentials;
 }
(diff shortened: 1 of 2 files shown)


packages/fetcher/src/IConnection.ts
```diff
@@ -47,4 +47,9 @@ export interface IConnection {
      * @default false
      */
     simulate?: boolean | Partial<IRandomGenerator>;
-}
+
+    /**
+     * A string indicating whether credentials will be sent with the request always, never, or only when sent to a same-origin URL. Sets request's credentials.
+     */
+    credentials?: RequestCredentials;
```
samchon: Change to string literal union type please

samchon (changes requested):
Thanks for contribution.

By the way, in NodeJS environment, the `RequestCredentials` type does no exist, so it would be compilation error. Therefore, you'd better to change it as string literal union type instead.

packages/fetcher/src/IConnection.ts
SeungjunWe: I replace it to string union type

samchon (approved):
Good, your PR will be published after testing

Issue #466, opened 2023-07-14
[Suggestion] add transformPath option for yarn workspace monorepo


SeungjunWe:
# Description
- https://github.com/samchon/nestia/issues/259

As shown in the issue above, the nestia/sdk interprets an import from another workspace in a monorepo as a relative path import.

This happens because the `Type` of the typescript gives an absolute path rather than the original import path, so the
It seems to be caused by the fact that the NESTIA SDK cannot distinguish whether the type is importing a package module or not.

For this reason, I've decided to create a 
`transformPath` option to tell the nestia sdk which path is a workspace module.

# What does it add?
You can add a transformPath when you have a monorepo folder structure like the one below.
```
- projectRoot
  - packages
    - dtoPackage
    - nestiaProject
```

nestia.config.ts
```
{
  // ...
  transformPath: {
    // 'worksapce module path': 'packageName'
    'packages/dtoPackage': '@workspace/dto',
  },
}
```

samchon:
I saw your repository, but it is not reproducible because of github private package.

By the way, this problem was not nestia, but of your `tsconfig.json` configuration. You're importing `@lee-kyungseok/blog-dto-sharing-sample-contract` from `node_modules`, but your yarn monorepo setting is guiding to `packages/contract` directory. To solve that problem, just configure `tsconfig.json` file to `include` only `src` and `node_modules` directories like below.

If problem still occurs even configured like that, then just tell me. I'll try another solution. Your PR cannot be accepted because it is a typical hard coding ignoring standard configuration way of `tsconfig.json''s `compilerOptions.paths` property.

```json
{
  "include": ["src", "node_modules"]
}
```

SeungjunWe:
> {
  "include": ["src", "node_modules"]
}

thanks for help, but it not works for me. T_T

samchon:
In my case, there's no problem even when constructing monorepo. Can you provide a reproducible repo?

PR #467, opened 2023-07-14
add transformPath option #466

SeungjunWe:
- #466


samchon (commented):
Reject due to violating TypeScript standard configuration method, `compilerOptions.paths`

PR #527, opened 2023-08-07
Fix samchon/typia#744 - change setup wizard to deprecate yarn berry

samchon:
Same with with samchon/typia#744

  - Changed setup wizard
  - Introduce not to support `yarn berry`
  - Use `npm run postinstall` command instead

Unfortunately, `ts-patch` does not support `yarn berry`. Therefore, it is not possible to use transform mode of `typia` in the `yarn berry`. Therefore, to inform it and recommend not to use `yarn berry` in `typia`, I've changed setup wizard program. This change would be rolled back when `ts-patch` starts supporting the `yarn berry`.

Also, `yarn berry` does not support the `npm run prepare` command. I think this nonsensible story because the `npm run prepare` command is a standard feature of NPM, but this is the reality, so I can't help it. By the way, if `ts-patch` starts supporting `yarn berry`, configuring `npm run prepare` command by setup wizard can be a big problem. So, I've changed setup wizard to use `npm run postinstall` command instead of the previous `npm run prepare` command.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "nestia",
-  "version": "4.3.2",
+  "version": "4.4.0",
   "description": "Nestia CLI tool",
   "main": "bin/index.js",
   "bin": {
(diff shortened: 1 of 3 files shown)


jsbrain:
I got nestia to work with berry in previous versions but now not anymore. I'm not quite sure if this was a good move.

You definitely need to use the `nodeLinker: node-modules` option and then it should work, at least for certain configurations.

Also as stated [here](https://github.com/nonara/ts-patch/issues/119) you might be able to utilize `yarn patch` which should work as well.

Issue #551, opened 2023-08-17
Provide the operation id


loucass003:
Please provide the operation id in the generated swagger document as it is required by some generators to work properly.
I will also mention that nestjs/swagger does specify the operation id

Here is what the spec says about it https://swagger.io/docs/specification/paths-and-operations/  and https://swagger.io/specification/v3/
``
Unique string used to identify the operation. The id MUST be unique among all operations described in the API. 
The operationId value is case-sensitive. 
Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is RECOMMENDED to follow common programming naming conventions.
``

For instance https://github.com/fabien0102/openapi-codegen/ requires the operationId to be set.

It looks like you can probably set the value to the same value as ``x-nestia-namespace`` from what i have read some generators use those operation id to name the generated code.

Here is a quick and dirty code that i made in attempt to temporarly fix this
```ts
  const methods = [
    'get',
    'put',
    'post',
    'delete',
    'options',
    'head',
    'patch',
    'trace',
  ];
  const camelize = (s) => s.replace(/-./g, (x) => x[1].toUpperCase());
  for (const [key] of Object.entries(swaggerDocument.paths)) {
    for (const method of methods) {
      if (swaggerDocument.paths[key][method]) {
      
        const name = swaggerDocument.paths[key][method]['x-nestia-namespace']
          .split('.')
          .map(camelize)
          .join('__');
        swaggerDocument.paths[key][method]['operationId'] = name;
      }
    }
  }
  SwaggerModule.setup('api', app, swaggerDocument, {
    swaggerOptions: {
      defaultModelRendering: 'model',
    },
  });
  ```
 
 Some nice to have would also to be able to set the operation id in the documentation like so
 
 
 ```ts
/**
     * Would be shown without any mark.
     *
     * @param section Section code
     * @param input Content to store
     * @returns Newly archived article
     * 
     * @tag public
     * @summary Public API
     * @security bearer
     * @security oauth2 read write
     * @operation myCustomNamedOperation
     */
    @TypedRoute.Post()
    public async store(
        @TypedParam("section") section: string,
        @TypedBody() input: IBbsArticle.IStore,
    ): Promise<IBbsArticle> {
        return {
            ...typia.random<IBbsArticle>(),
            ...input,
            section,
        };
    }     
 ```
 
 this would require some validation before generating the schema as you would have to ensure that he operationId are unique

samchon:
Upgrade to the latest version, then it would be possible.

PR #562, opened 2023-08-26
Adding the operation id

loucass003:
Here is my atempt at adding the operationid based on https://github.com/samchon/nestia/issues/551

Schema is generating fine. i just have no clue on how to implement a test for it


samchon:
Will review and accept this with `typia@v5` and `@nestia/fetcher` update.

loucass003:
Why removing the generated default operationId?

The mail goal of this pr was to have that generated operation id, so code generators could use it as an unique key.
Removing that generated operation id mean that i will have to manually specify it on all controller method.
it is to the generator to decide how to use it, but we should imo always send something.

samchon:
Got it, then I will add a callback function determining operationId on nestia.config.ts file.

Issue #563, opened 2023-08-27
Ability to set the typescript config file path for sdk generation


dragonol:
So, I have 2 typescript config files: `tsconfig.json` and `tsconfig.app.json`. And when I'm trying to run `nestia sdk`, look like the command always use the `tsconfig.json` config while I want to use `tsconfig.app.json` config for the sdk generation. Is there any way to do that? Thanks

samchon:
I'm developing `typia` feartures in nowadays, so this idea seems good, but I can't do it right now.

If you can't wait, it is okay to sending a PR.

If not, wait one or two weeks please.

samchon:
Upgrade to v2, then be come true.

samchon:
@dragonol  I've started supporting this feature, and it can't be compatible with this `project` property.

Therefore, decided to support your suggestion by CLI way like below since v2.3.1 update.

Sorry for sudden break change.

```bash
npx nestia sdk --project nestia.tsconfig.json
```

PR #565, opened 2023-08-27
feat: ability to customize tsconfig path

dragonol:
## Issue #563 

I tested on my own project, but I cannot find any existing unit test in sdk package to lookup. Any guide for me to follow?

### DIFF
--- a/packages/sdk/src/INestiaConfig.ts
+++ b/packages/sdk/src/INestiaConfig.ts
@@ -135,6 +135,13 @@ export interface INestiaConfig {
      * @default false
      */
     simulate?: boolean;
+
+    /**
+     * Project configuration
+     *
+     * Path to typescript configuration file, or to a folder with a 'tsconfig.json'
+     */
+    project?: string;
 }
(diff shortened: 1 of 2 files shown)


samchon:
It is under test directory. Move to there and make a new project as you wish.

By the way, as every projects under the test directory are considered to using only tsconfig.json file, it may need to change test/executable/start.js file a little bit different. Id it seems hard, I will do it at Wednesday.

Thanks for contribution.

```bash
cd test
npm run generate title

 # after completion of the project
npm start -- --only title

# when library be changed
npm run clean
npm run start
```

samchon (changes requested):
By the way, the word project would be better, because of tsc command like tsc --project configFile command

dragonol:
Just updated the wording, thanks @samchon

samchon:
Will accept it with `typia@v5` and `@nestia/fetcher` update.

samchon (approved):
Thanks for contribution.

I'll adjust your suggestion to v2 update.

PR #584, opened 2023-09-03
Complement #551 - callback function for `operationId` of swagger.

samchon:
Added callback function configuring default `operationId` of swagger in `netia.config.ts` file. However, to support this feature, I decided to discard CLI generation mode of SDK and swagger documents.

@loucass003 - please review this concepts. Also, I wanna ask you that default must be configured `operationId` even both `@operationId` and callback function are omitted?

https://github.com/samchon/nestia/blob/8562f286ab207efbad6ba8721e97a0ca63c196ba/packages/sdk/src/INestiaConfig.ts#L175-L180


loucass003:
Sorry for the late response, is my review still required? as this pr is already merged

PR #585, opened 2023-09-03
Complement #563 - finalize `INestiaConfig.project`

samchon:
When multiple `tsconfig.json` files exists, it is possible to specify one thing by configurint `INestiaConfig.project` property of `nestia.config.ts` file.


samchon (commented):
Test passed in local

PR #588, opened 2023-09-04
Add swagger tag from controler doc comment

trangianhuan:
# Add swagger tag from controler doc comment

ảnh

ảnh


samchon (commented):
What about using @ApiTags() decorator instead? I will support it at next v2 update.

Your suggestion seems good, but the reason I am not parsing any comment tags from class. It is because if a class being extended, lots of contradicts being occured from the inheritance. Sorry for rejecting in such reason.

PR #617, opened 2023-09-19
Fix(typo): Update setup.mdx

dipanshuhappy:
In the setup.mdx , it said 
`When you want to set up nestia in exiting project ....` it should say 
`existing` rather than `exiting`


samchon (approved):
Thanks for emension

PR #642, opened 2023-10-02
fix: resolved path might be pattern in glob

YJ-AnthonyJo:
### When
- project's base path(rootDir) has string which matches to Glob Pattern(eg. [], *)
- generating swagger or sdk based on `input` paths of `nestia.config.ts`

My case, My base path is `~/dev/[project]inprogress/ProjectName`  
image

### Expected Behavior
- sdk should find `.ts`files inside the paths of `input` of `nestia.config.ts`, analyze those `.ts` files and display the results to console properly.

image

### Actual Behavior
- sdk finds the wrong files because it considers the string of the base path (matching the Globe pattern) to be a Globe pattern.

My case, It couldn't find any files.
image

### How to reproduce
- Make nestjs and nestia project inside `[test]nestia`

## How did I fix
Replaced `path.resolve(pattern)` with `pattern` in `emplace` arrow function of `utils/SourceFinder.ts`

I think this change also serve its original purpose as before.
- `nestia.config.ts` is on root directory.
- `glob(pattern)` method acts on relative path so `path.resolve(pattern)` is unnecessary. (Only the difference between relative or absolute paths.)
- In `_Glob` method, path is resolved when they match so it can be used to subsequent works
image

image

## Enviroment & Setup
- OS: macOS Ventura 13.3.1
- node version: 20.4.0
- nestjs version: `@nestia/sdk` 2.0.5

### DIFF
--- a/packages/sdk/src/utils/SourceFinder.ts
+++ b/packages/sdk/src/utils/SourceFinder.ts
@@ -20,7 +20,7 @@ export namespace SourceFinder {
         (input: string[]) =>
         async (closure: (location: string) => void): Promise<void> => {
             for (const pattern of input) {
-                for (const file of await _Glob(path.resolve(pattern))) {
+                for (const file of await _Glob(pattern)) {
                     const stats: fs.Stats = await fs.promises.stat(file);
                     if (stats.isDirectory() === true)
                         await iterate(filter)(closure)(file);


samchon (approved):
OK, I'll publish the next version 2.1.7

PR #644, opened 2023-10-08
Add new decorator `@TypedQuery.Body()`

samchon:
To support `application/x-www-form-urlencoded` content-typed request body, added new decorator `@TypedQuery.Body()`.

About the response body case, it would be supported by the next PR.


samchon (commented):
OK, succeeded to pass the test program at the 1st trial.

PR #650, opened 2023-10-10
Publish v2.2 with `@nestia/migrate`

samchon:
Have published the v2.2 update.


samchon (commented):
There would not be any problem. Just merge direclty.

PR #659, opened 2023-10-14
Close #656 -support multiple nestia.config.ts file.

### LINKED ISSUE #656: Support multiple `nestia.config.ts` file
```bash
npx nestia sdk --config nestia.config.ts
npx nestia swagger --config something.config.ts
```

As `@nestia/sdk` has started supporting the propagation mode (#599), I think there can be some users who want to build SDK in both normal and propagation mode at the same time. To support such future requirement, I've decided to support multiple `nestia.config.ts` files by adding a new parameter `--config` in the CLI command.

samchon:
```bash
npx nestia sdk --config nestia.config.ts
npx nestia swagger --config something.config.ts
```

As `@nestia/sdk` has started supporting the propagation mode (#599), I think there can be some users who want to build SDK in both normal and propagation mode at the same time. To support such future requirement, I've decided to support multiple `nestia.config.ts` files by adding a new parameter `--config` in the CLI command.


samchon (commented):
Succeeded to pass the test programs.

PR #661, opened 2023-10-18
Supports `INestApplication` on `nestia.config.ts` file

samchon:
```typescript
import { INestiaConfig } from "@nestia/sdk";
import { Module, VersioningType } from "@nestjs/common";
import { NestFactory } from "@nestjs/core";

    input: async () => {
        const app = await NestFactory.create(MyModule);
        app.setGlobalPrefix("api");
        app.enableVersioning({
            type: VersioningType.URI,
            prefix: "v",
        });
        return app;
    },
    output: "src/api",
    e2e: "src/test",
    swagger: {
        output: "swagger.json",
        security: {
            bearer: {
                type: "apiKey",
            },
        },
    },
};
export default NESTIA_CONFIG;
```

From now on, you can configure `INestApplication` instance to the `INestiaConfig.input` property. 

So, you can utilize special features of NestJS like [`setGlobalPrefix()`](https://docs.nestjs.com/faq/global-prefix), [`enableVersioning()`](https://docs.nestjs.com/techniques/versioning), and [`RouterModule`](https://docs.nestjs.com/recipes/router-module).

@micalevisk Can you review this approach is proper? I've succeeded to pass the test program of `nestia`, but I cannot sure this code works in every case because the `any` type castings are too much used.

https://github.com/samchon/nestia/pull/661/files#diff-c8a57828e08f6f3a74e547bdfb072b31f146e646345a636642ccb332a7ca01c6R36-R102


samchon (commented):
Succeeded to pass all test programs, and there seems not any problem about the module input.

PR #668, opened 2023-10-19
Report errors during SDK generation.

samchon:
When an error of NestJS project being detected in the SDK generation process, `@nestia/sdk` had thrown an exception. Therefore, if there're many errors in the NestJS project, user may lost many times due to only one error being reported.

To fix the problem, I've updated `@nestia/sdk` to gather every errors of target NestJS project, and print all of them at once.


samchon (commented):
Succeeded to pass every test programs in local.

Github actions may pass too.

PR #674, opened 2023-10-24
fix: Typo in sdk docs - missing status code 200

ninthsun91:
A minor typo fix in Propagation description.

image

I suppose status code `200` is missing after *"not"*


samchon (approved):
Thanks for emension.

PR #683, opened 2023-11-08
`ts.CompilerHost.jsDocParsingMode`

samchon:
Since TypeScript v5.3 update, TypeScript compiler no more analyzes `JsDocComment` as default.

> The desire of that is to boosting up the compilation speed by not parsing the JS Doc Comments.

Therefore, to parepare the future TypeScript v5.3 release, I've updated `@nestia/sdk` to configure the `jsDocParsingMode` to parse every comments manually. For reference, the restriction of `peerDependencies` -> `"typescript": ">=4.8.0 <5.3.0"` would also be erased after formal TS v5.3 release.

https://github.com/microsoft/TypeScript/pull/55739

### DIFF
--- a/benchmark/package.json
+++ b/benchmark/package.json
@@ -25,7 +25,7 @@
   "homepage": "https://nestia.io",
   "dependencies": {
     "@nestia/core": "^2.2.1-dev.20231012",
-    "typia": "^5.2.5"
+    "typia": "^5.2.6"
   },
   "devDependencies": {
     "@trivago/prettier-plugin-sort-imports": "^4.1.1",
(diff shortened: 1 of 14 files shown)


jakebailey:
The API will still parse JSDoc by default; only tsc has this disabled (and it's likely ts-patch will undo that change). You shouldn't need to do anything on your side.

samchon:
> The API will still parse JSDoc by default; only tsc has this disabled (and it's likely ts-patch will undo that change). You shouldn't need to do anything on your side.

Ah, got it. Thanks for detailed info.

PR #684, opened 2023-11-08
Fix #680 - when target type of `TypedException<T>` is union and shuffled

### LINKED ISSUE #680: [Bug Report] TypedException<T> does not work conditionally
# Bug Report
when Exception Type is Union type, type is missing conditionally.

## Bug Condition
1. define api using TypedException
2. in another controller (in another file) compiled later, define union exception type including step 1 type.
3. when build sdk or swagger, step 2 exception info is missing.

## Solution
In nestia/sdk/analyses/ExceptionAnalyzer `DO ASSIGN` code, exception info is filtering.

### Orignal Code
```ts
// DO ASSIGN
const matched: IController.IException[] = Object.entries(func.exceptions)
                .filter(([_key, value]) => value.type === tuple.typeName)
                .map(([_key, value]) => value);
```
when tuple typeName is "B" | "A" | "C" and value.type is "A" | "B" | "C", this case is filtered.

### Fixed Code
```ts
// DO ASSIGN
const matched: IController.IException[] = Object.entries(func.exceptions)
                .filter(([_key, value]) =>
                    value.type.includes(" | ") && tuple.typeName.includes(" | ")
                        ? value.type.split(" | ").sort().join(" | ") === tuple.typeName.split(" | ").sort().join(" | ")
                        : value.type === tuple.typeName,
                )
                .map(([_key, value]) => value);
```

industriously:

```bash
# COMPILE THE BACKEND SERVER
npm run build

npm run test
```


samchon (approved):
Great point, and nice test code.

PR #692, opened 2023-11-23
Fix samchon/typia#880 - `any` type comes when `assertClone` option.

samchon:
In nowadays, I've added an option that configure `assertClone` to be `TypedBody()`'s validator. 

By the way, I've taken a mistake that have not considered the any typed property case from the request body DTO.

This PR fixes the problem, by supporting the `any` typed value cloning.


samchon (commented):
No problem on the test program, so commit it.

PR #695, opened 2023-11-25
Close #693 - `@ignore` comment tag.

### LINKED ISSUE #693: Add @Ignore decorator to exclude endpoints from sdk

samchon:
If you want to exclude a controller method from SDK library, write `@ignore` comment on the target controller method.

Then SDK generator will ignore it.


samchon (commented):
No problem in the test programs.

PR #702, opened 2023-11-28
Add `beautify` option for swagger generation.

samchon:
Until now, `@nestia/sdk` always beautified `swagger.json` content. However, some user requested me to give an option to minify or beautify when generating the `swagger.json` file, so that I've newly added an option `INestiaConfig.ISwaggerConfig.beautify` like below:

```typescript
export interface INestiaConfig {
    swagger: INestiaConfig.ISwaggerConfig;
}
export namespace INestiaConfig {
    export interface ISwaggerConfig {
        /**
         * Whether to beautify JSON content or not.
         *
         * If you configure this property to be `true`, the `swagger.json` file would
         * be beautified with indentation (2 spaces) and line breaks. If you configure
         * numeric value instead, the indentation would be specified by the number.
         *
         * @default false
         */
        beautify?: boolean | number;
    }
}
```


samchon (commented):
No problem from test programs

PR #716, opened 2023-12-07
Complement `EncrypedBody()` and `@EncrypedRoute()`.

samchon:
Tried to change `EncryptedBody()` and `@EncryptedRoute()` to utiltize binary data instead of base64 string, but failed due to default `FastifyAdaptor` does not support the `application/octet-stream`.

Instead, fixed those decorators to using the base64 encoding with `text/plain` content type, and complemented swagger document's mis-writing description field problem.


samchon (commented):
No problem on testing.

PR #739, opened 2024-01-09
Make swagger file lighten for LLM reason + invalid path checking

samchon:
Same reason with https://github.com/samchon/typia/pull/916

Also, when invalid path like `::id` comes, SDK/swagger generator will print the error with reason why instead of terminating the generator program throwing exception message.


samchon (commented):
Suceeded to pass every test programs.

PR #768, opened 2024-01-26
Adapt new format tags of `typia`.

samchon:
```typescript
export interface TypeTagFormat {
  // SPECIAL CHARACTERS
  byte: string & tags.Format<"byte">;
  password: string & tags.Format<"password">;
  regex: string & tags.Format<"regex">;
  uuid: string & tags.Format<"uuid">;

  // ADDRESSES
  email: string & tags.Format<"email">;
  hostname: string & tags.Format<"hostname">;
  idnEmail: string & tags.Format<"idn-email">;
  idnHostname: string & tags.Format<"idn-hostname">;
  iri: string & tags.Format<"iri">;
  iriReference: string & tags.Format<"iri-reference">;
  ipv4: string & tags.Format<"ipv4">;
  ipv6: string & tags.Format<"ipv6">;
  uri: string & tags.Format<"uri">;
  uriReference: string & tags.Format<"uri-reference">;
  uriTemplate: string & tags.Format<"uri-template">;
  url: string & tags.Format<"url">;

  // TIMESTAMPS
  datetime: string & tags.Format<"date-time">;
  date: string & tags.Format<"date">;
  time: string & tags.Format<"time">;
  duration: string & tags.Format<"duration">;

  // POINTERS
  jsonPointer: string & tags.Format<"json-pointer">;
  relativeJsonPointer: string & tags.Format<"relative-json-pointer">;
}
```

https://github.com/samchon/typia/releases/tag/v5.4.1

`typia` has started supporting new `format` tags of `string` type. Also, in the release, regular expression of `url` tag has been changed. To adapt and encourage the new format tags of `string` type like `Format<"iri-reference">`, `nestia` also releases new patch with `typia` dependency updating.

Also, remove `typescript` from `peerDependencies` in the `@nestia/core` and `@nestia/sdk` modules, because dependent library `typia` already has the same property. Therefore, as `typia` can limit the TypeScript version range, `nestia` does not need to check the TypeScript version duplicatedly.


samchon (commented):
Everything works fine.

PR #771, opened 2024-01-29
Support OpenAPI v3.1

samchon:
Configure `swagger.openapi` property value to be `"3.1"` in the `nestia.config.ts` file.

Then, `@nestia/sdk` will generate OpenAPI v3.1 document instead of v3.0.

```typescript
import { INestiaConfig } from "@nestia/sdk";

export const NESTIA_CONFIG: INestiaConfig = {
  input: ["src/controllers"],
  output: "src/api",
  e2e: "src/test",
  swagger: {
    openapi: "3.1",
    output: "swagger.json",
    security: {
      bearer: {
        type: "apiKey",
      },
    },
    beautify: true,
  },
};
export default NESTIA_CONFIG;
```


samchon (commented):
As `typia` is not fully supporting OpenAPI v3.1 spec's JSON schema, it needs to be a little bit delayed.

PR #770, opened 2024-01-28
Close #769 - SDK library generation through compiler API

samchon:
Until now, I've utilized string concat hard coding when generating SDK library.

Although there had been not any problem until now, it is hard to maintaining and be an obstacle for other contributors. Furthremore, such strinng concat hard coding will be a great technical debt when more features or enhancement being required in the `@nestia/sdk` library.

Therefore, this PR has changed the SDK generator to utilize not string concat based hard coding, but TypeScript compiler API. Also, as TypeScript compiler API does not support `span` feature that adding new line or formatting like `pretter`, I've changed the SDK generator to call the `prettier` API function for the formatting.

Issue #682, opened 2023-11-08
[Suggestion] build sdk only using swagger docs


2langk:
## Feature Request
I wrote swagger document using spring boot.
and I want to build nestia sdk using swagger document. 
now, it is impossible to build only sdk using `@nestia/migrate` (it can generate sdk but also nestjs application)

so, I suggest options like `--sdk-only`  in `@nestia/migrate`

example)
`npx @nestia/migrate --sdk-only swagger.json output_directory`

samchon:
https://github.com/samchon/nestia/issues/445

Good suggestion.

To accomplish above feature `@nestia/editor`, I should implement this issue, and it was one of my plan.

samchon:
@2langk 

Completed 

https://nestia.io/docs/migrate/

PR #773, opened 2024-01-30
Also use compiler API in `@nestia/migrate`.

samchon:
Like #770 case that SDK library utilizes TypeScript compiler API instead of using the string concat hard coding, `@nestia/migrate` also did it.


samchon (commented):
Next step is #682, and need to change CLI grammer.

samchon (commented):
Succeeded to pass every test programs.

PR #775, opened 2024-01-31
Close #682 - SDK generation from Swagger Documents

### LINKED ISSUE #682: [Suggestion] build sdk only using swagger docs

samchon:
From now on, `@nestia/migrate` can make SDK library by itself.

By the way, have to consider how to compose the CLI command about that feature.


samchon (commented):
Developed new CLI command and also inquirer when CLI be missed

PR #798, opened 2024-02-15
fix nestia core dependencies

AngPuoSyer:
### Main changes
moved `peerDependencies` that is in `dependencies` into `devDependencies` and remove unnecessary peerDependencies.

### Issues
The dependencies that should be in peerDependencies are causing peer dependency conflict with older version of NestJs

#### Dependencies moved to devDependencies
- @nestia/fetcher,
- @nestjs/common
- @nestjs/core
- raw-body
- reflect-metadata
- rxjs
- typia

#### Dependencies removed
- @nestjs/platform-express
- @nestjs/platform-fastify


samchon:
How older version of NestJS you are using?

The libraries you've erased in the `dependencies` are actually used in the `@nestia/core`. 

If you remove them, there would not be any problem when you're using `npm`, but becomes error in the `pnpm`.

samchon (approved):
On testing, thanks for good suggestion.

samchon:
Will be published after 2 hours later with deep testings.

samchon:
The dependency problem had not been caused by the `dependencies` of `@nestia/core`, but from `@nestjs/platform-express` and `@nestjs/platform-fastify`. I'll fix it after an hour.

PR #815, opened 2024-02-26
Refactor `@nestia/migrate` for 3rd party application.

samchon:
`ISwaggerComponents.namespace` property is for 3rd party application supporting.


samchon (commented):
No problem.

PR #833, opened 2024-03-15
feat: add wrapper parameters

rojiwon123:
in last version, nestia e2e wrapper can't use options.parameters function (but parameters function is required)

so, it is more good that wrapper function get options.parameters at 3rd function parameters.


samchon (approved):
Thanks for contribution.

PR #843, opened 2024-03-24
Add new property `IConnection.logger`.

samchon:


samchon (commented):
Succeeded to pass every test programs.

PR #847, opened 2024-03-25
Feat: Support Swagger Optional Security Tag

rojiwon123:
In OpenAPI, optional security is supported, which refers to APIs where authorization tokens may or may not be present. To support this in Nestia, some code modifications were made.

Below is how you can add optional security:

```ts
@nest.Controller("security")
export class Controller{
  /**
  * @security
  * @security bearer
  */
  async security(){}
}

```
about #849
about https://github.com/OAI/OpenAPI-Specification/issues/14


samchon:
Upgrade to `v2.6.1`, then your feature be activated.

PR #851, opened 2024-03-27
Fix: incorrect code example order in TypedBody docs

koh1260:
## Feature
- Modifying the order of the tag list in the How To Use section of the TypeBody.

## Before
스크린샷 2024-03-27 오후 7 21 59

## After
스크린샷 2024-03-27 오후 7 21 41


samchon (commented):
It is intended order \o/

PR #907, opened 2024-05-24
benchmark: Apple M1

sytranvn:


samchon (approved):
Thanks for your measurement.

PR #960, opened 2024-07-15
fix/docs: replace FastifyAdaptor with FastifyAdapter

LeUKi:
`FastifyAdapter` is a valid export of @nestjs/platform-fastify, which is described as `FastifyAdaptor` in the comments of the template in nestia.config.ts, which is confusing for newbies, this PR fixes all the wrong exports and uses of `FastifyAdaptor`.
[https://docs.nestjs.com/techniques/performance](https://docs.nestjs.com/techniques/performance)


samchon (approved):
Oops, thanks for taking making my mistakes.

PR #968, opened 2024-07-24
Fix #966: support `jsx` configured case.

### LINKED ISSUE #966: npx nestia sdk -> error TS5024: Compiler option 'jsx' requires a value of type string.
When running the `npx nestia sdk` command, I receive the following error:
`error TS5024: Compiler option 'jsx' requires a value of type string.`

i have a workaround by removing `"jsx": "preserve",` from my tsconfig.json, running the command, then adding the line back. Tho this isnt ideal.

I'm not sure if this is a Nestia related issue or if its typescript itself

Step by step
1. jsx option is present
image
2. command fails
image
3. remove jsx option
image
4. command all good
image
5. add jsx option back to tsconfig

p-sw:
Here is the issue: #966 

Running nestia sdk command with tsconfig compilerOption `jsx` cause the error below:
```
error TS5024: Compiler option 'jsx' requires a value of type string.
```

## Reason
1. `nestia sdk` internally uses NestiaConfigLoader
2. NestiaConfigLoader's which uses `tsconfck.parseNative` to parse compilerOption from tsconfig (reference [here](https://github.com/samchon/nestia/blob/master/packages/sdk/src/executable/internal/NestiaConfigLoader.ts#L20))
3. `tsconfck.parseNative` uses typescript's native function, `readConfigFile` and `parseJsonConfigFileContent`
4. `parseJsonConfigFileContent` causes error - it resolves `jsx` to `number`, but `jsx` should be string!
5. ERROR

## Solution in this PR
* Replace `tsconfck.parseNative` to `tsconfck.parse`

## Other changes
* Bump tsconfck to ^3.1.1

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -42,7 +42,7 @@
     "path-to-regexp": "^6.2.1",
     "prettier": "^3.2.5",
     "ts-node": ">=10.6.0",
-    "tsconfck": "^3.1.1",
+    "tsconfck": "^2.1.2",
     "tsconfig-paths": "^4.1.1",
     "tstl": "^3.0.0",
     "typia": "^6.5.4"
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -42,7 +42,7 @@
     "path-to-regexp": "^6.2.1",
     "prettier": "^3.2.5",
     "ts-node": ">=10.6.0",
-    "tsconfck": "^2.0.1",
+    "tsconfck": "^3.1.1",
     "tsconfig-paths": "^4.1.1",
     "tstl": "^3.0.0",
     "typia": "^6.5.4"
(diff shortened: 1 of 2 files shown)


p-sw:
It seems like the test is failing because of working directory.

I run `npm run test` on the root directory, it works fine without any error.

But If I `cd test` and then `npm run test`, it cause the error same as action.

I'm not sure why...

samchon:
@p-sw The test program is running the `npx nestia sdk` or `npx nestia swagger` command in each directory by configuring specific `cwd (current working directory)`.

samchon (changes requested):
@p-sw The test program is running the `npx nestia sdk` or `npx nestia swagger` command in each directory by configuring specific `cwd (current working directory)`.

The bug has come from bumping up the `tsconfck` dependency, which does not support CJS program.

samchon (commented):
Rolling back the `tsconfck` version will make your PR work.

Issue #980, opened 2024-08-07
Boost up SDK and Swagger Document generation speed


samchon:
Current `@nestia/sdk` is performing `dynamic import` to every controller files inidividually with `ts-node` even though NestJS application module already being mounted in the `ts-node` process. This step had been essentially required due to reflection information acquisition and matching the reflection information with metadata of each API operations.

By the way, as `dynamic import` statement in `ts-node` occurs excessive load on the TypeScript compiler, SDK and Swagger Document generation was slower. For example, my example backend server has 200 number of API endpoints, and its consumption time on SDK or Swagger Document generation overs 30 seconds.

In today, I got an idea to reduce the SDK and Swagger Document generation time shorter. It is to making an internal transformer into the `@nestia/sdk`, and the internal transformer of `@nestia/sdk` adjusts a method decorator with metadata information. The metadata information would be `ts.Type` and description comment data.

With this strategy, `@nestia/sdk` still needs `ts-node`, but does not need to inidividual `dynamic import` step. The metadata of API operation would be enrolled to the metadata, so I expect that SDK and Swagger Document generation time would be about 10x faster.

To accomplish this mission, the transformed raw level (JSON level) metadata must be delivered to the runtime, and the runtime must revive the metadata. However, `ts.Type` is a type of class with a lot of methods, and its member properties are recursive. If the transformer writes the primitive literal object value (JSON value) of metadata to the JavaScript, its recursive member properties and methods would be broken.

Therefore, I'm testing serializing and derializing such `ts.Type` class instance in both compilation and runtime level. My experiment is going on `typia` library. If my experiment concludes to failure, this issue would be closed. Otherwise, succeess to reviving the `ts.Type` class instance, the generation performance would be dramatically enhanced.

samchon:
Suceeded to revive `ts.Type`, but cannot sure it would work on generator.

In tomorrow, I'll test it.

samchon:
Failed to de/serialize experiment in `typia`, but got another idea.

It is to composing a transformer into the `@nestia/sdk` and getting JSON schema and type name ealirer then generation.

samchon:
Completed.

Even though the SDK and Swagger Document generation speeds are not significantly faster than before when building only one module, it would be dramatically boosted up when building multiple swagger & SDK library files/modules.

PR #982, opened 2024-08-10
Developing #980: to boost up `@nestia/sdk` performance.

samchon:


samchon (commented):
Finally succeeded to pass every test programs.

PR #985, opened 2024-08-14
Allow multiple configurations in a file.

samchon:


samchon (commented):
Everything is fine, @dependabot merge

PR #987, opened 2024-08-14
Fix request body content type bug of SDK library.

samchon:
As `@nestia/sdk@3.11.0` generated SDK library always filled request content body type even when request body does not exist. By the phenomenon, it sometimes had caused HTTP request bug in POST/PATCH/PUT/DELETE methoded requests when request body does not exist.

This PR fixes the bug, just by filling the request body's content type only when it is required.


samchon (commented):
No problem more.

PR #989, opened 2024-08-14
Test programs' re-compilations

samchon:
As `@nestia/sdk` has been entirely remade in v3.11 update, entirely re-compiled test programs too.


samchon (commented):
OK, fixed mis-merged problem.

PR #997, opened 2024-08-22
Avoid reserved words in `@nestia/migrate`.

samchon:
Looking at github's swagger.json file, reserved keyword like `public`, `private` are used in `OpenApi.IOperation.path` and type name in `OpenApi.IJsonSchema.IReference.$ref`.

No way at type name, but it would better to avoid the reserved keyword on `IMigrateRoute.accessor` composed by parsing and spitting the `OpenApi.IOperation.path`.

This PR fixes the problem, so that helps OpenAPI generator libraries to be safer.


samchon (commented):
OK just proceed it.

PR #1010, opened 2024-09-03
Close #1007: `NoTransformConfigurationError.throws`

### LINKED ISSUE #1007: Option to disable nestia in tests [$$$]
## Feature Request

I can contribute financially to implement this feature request.

  - A description of the problem you're trying to solve.
  
My Jest tests finish fairly quickly when using `ts-jest` with the `isolatedModules` flag enabled. However, nestia won't work with that flag because (I assume) it looks for type information which isn't there.

It raises the following error:
`TypeError: Cannot read properties of undefined (reading '0')`
  
When I disable the `isolatedModules` flag, my tests pass but their execution time becomes way too high. It increases almost ten-fold.
  
  - An overview of the suggested solution.
  
I would like to be able to run my Jest tests with the `isolatedModules` flag enabled. To achieve that, I need to be able to disable nestia run-time validation. I can do that by using a dedicated tsconfig file without the `plugins` parameter. Obviously, it makes nestia complain that `no transform has been configured.`

If nestia accepted an explicit `null` driver or something of that kind, I would be able to run my tests with the mentioned flag. The obvious drawback is that the run-time validation would be gone. But that's not something I want to test so that doesn't concern me.

I really appreciate nestia. As mentioned, I can contribute financially if this feature request is honored and consequently, I'm able to use the library in my project.

samchon:
Disable no transformation error by configuring `NoTransformationError.throws = false`.

```typescript
import { NoTransformConfigurationError } from "@nestia/core/lib/decorators/NoTransformConfigurationError";
import { DynamicExecutor } from "@nestia/e2e";

// prettier-ignore
NoTransformConfigurationError.throws = false;

// prettier-ignore
import { Backend } from "../Backend";

async function main(): Promise<void> {
  const server: Backend = new Backend();
  await server.open();

  const report: DynamicExecutor.IReport = await DynamicExecutor.validate({
    extension: __filename.substring(__filename.length - 2),
    prefix: "test",
    parameters: () => [
      {
        host: "http://127.0.0.1:37000",
        encryption: {
          key: "A".repeat(32),
          iv: "B".repeat(16),
        },
      },
    ],
    location: `${__dirname}/features`,
    onComplete: (exec) => {
      if (exec.error === null) {
      const elapsed: number =
        new Date(exec.completed_at).getTime() -
        new Date(exec.started_at).getTime();
      console.log(`  - ${exec.name}: ${elapsed.toLocaleString()} ms`);
      } else
        console.log(`  - ${exec.name}: ${exec.error.name}`);
    },
  });
  await server.close();

  const exceptions: Error[] = report.executions
    .filter((exec) => exec.error !== null)
    .map((exec) => exec.error!);
  if (exceptions.length === 0) {
    console.log("Success");
    console.log("Elapsed time", report.time.toLocaleString(), `ms`);
  } else {
    for (const exp of exceptions) console.log(exp);
    console.log("Failed");
    console.log("Elapsed time", report.time.toLocaleString(), `ms`);
    process.exit(-1);
  }
}
main().catch((exp) => {
  console.log(exp);
  process.exit(-1);
});
```


samchon (commented):
Test passed.

PR #1017, opened 2024-09-05
Test: Add `clone type create duplicate` test case

rojiwon123:
This PR cannot pass the tests because it includes test cases for the current bug in Nestia.


samchon (approved):
OK, I'll test and try to fix it now.

PR #1020, opened 2024-09-06
Update GenericAnalyzer.ts

shahriar-shojib:
I was getting this error when generating swagger document:
```
TypeError: Cannot read properties of undefined (reading 'getDeclarations')
```

When I added null check to genericAnalyzer.ts I was able to get it working on my local project.


samchon (approved):
Thanks for contribution. Would be published to v3.12.3 after testing.

PR #1021, opened 2024-09-10
Package manager detection like samchon/typia#1245.

samchon:
Contributed feature by @ryoppippi from `typia` to here `nestia`.


samchon (commented):
No problem.

PR #1033, opened 2024-09-17
Fix #1018: add `FormDataInput<T>` type for React Native environment.

### LINKED ISSUE #1018: Add support for React Native in @TypedFormData.Body()
## Feature Request

In react native the only way to send Files in a multipart request is by doing the following : 
```jsx
const formdata = new FormData();
formdata.append('file', {
  uri: 'path-to-file-on-the-phone',
  name: 'file-name',
  type: 'image/png',
});
```
this doesn't work : 
```javascript
formdata.append('file', new Blob(...))
formdata.append('file', new File(...))
```

As expected when using the first approach I will get the following error when using the custom object:
```
[Error: {"path":"$input.file","reason":"Error on typia.http.assertFormData(): invalid type on $input.file, expect to be Blob","expected":"Blob","value":"[object Object]","message":"Request multipart data is not following the promised type."}]
```

I think that to support all environments, validation inside ```@TypedFormData.Body()``` should support the custom object as a valid Blob because, at least, in fastify the custom object works the same way as a Blob without any additional handling needed
```javascript
//this is not of type Blob but it works when sent to fastify
{
  uri: 'path-to-file-on-the-phone',
  name: 'file-name',
  type: 'image/png',
}
```

samchon:
In the React Native, it does not support the `File` class.

Instead, it can compose the `FormData` class instance by appending below primitive object type. The primitive type can replace the `File` class instance in the React Native.

In such reason, added `FormDataInput<T>` type which casts `File` typed property value as `File | IFileProps` union type, and changed SDK generator to utilize the type for the `multipart/form-data` request body type.


samchon (commented):
OK, no problem.

PR #1051, opened 2024-09-26
Close #1037: validate.log option for @TypedRoute decorators.

### LINKED ISSUE #1037: I want to log errors that response type is wrong when `pugins.stringify` is null in `tsconfig.json`
## Feature Request

- currently when `plugins.stringify` in tsconfig.json is `null`, the response validation is not worked. but I want to log about error that explains "Invalid response".
  - what I want is that I don't want to validate the response, just take an error log that explains "Invalid response"

```typescript
// tsconfig.json
...
compilerOptions: {
  "plugins": [
      {
        ...
        "stringify": "log", // when use "log" type, nestia doesn't validate response type, but log the error message.
        // or
        "stringify": "custom", // when use "custom" type, user can customize about the response data. for example, validate response type and apply TransformInterceptor to response data using `@TypedRoute.Method("path", callback)`
      }
    ]
}

```

samchon:


samchon (commented):
Every tests succeded

PR #1053, opened 2024-09-26
Allow only `validate.log` option and use `typia.json.validateStringify()`.

samchon:
Erased validateEquals.log option, and changed the validate.log to use typia.json.validateStringify() for compact JSON serialization.


samchon (commented):
Okay

PR #1092, opened 2024-10-30
Avoid OpenAPI operation by `@hidden` comment tag.

samchon:


samchon (commented):
Local test OK, github actions also may okay.

PR #1103, opened 2024-11-11
Fix #1094: skip validation when no request body

### LINKED ISSUE #1094: Skip validation of content-type by TypedBody if body is empty
## Feature Request
In some cases I'd like to be able to make the body of a request _optional_. This seems to work as expected with `@TypedBody`, with the exception of it enforcing a check for a `Content-Type` of `application/json`:

https://github.com/samchon/nestia/blob/807474a7c7220922822e339f0c7d0dcd6d858491/packages/core/src/decorators/TypedBody.ts#L38-L41

I would argue that specifically in the case of a request with no body, there should be no expectation of a `Content-Type` header and therefore we should bypass this check. It would then be up to the validator to make a determination of whether this empty body is permissible.

samchon:


samchon (commented):
Everything is prepared.

PR #1130, opened 2024-12-04
Fix samchon/openapi#104 in `migrate` and `editor`.

samchon:
Related PR in `@samchon/openapi`: samchon/openapi#105


samchon (commented):
OK, tested in local.

PR #1153, opened 2024-12-13
improve node management

ryoppippi:
- manage pnpm version by pnpm
- manage node version by ci
- install node & pnpm using pnpm itself in ci env
-

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -12,17 +12,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 20.x
+
       - uses: pnpm/action-setup@v2
-        with:
-          version: 8
       - name: Root Setup
-        run: pnpm install && npm run build
+        run: |
+          pnpm install
+          pnpm run build
       - name: Build
         working-directory: website 
-        run: npm install && npm run build
+        run: |
+          pnpm install
+          pnpm run build
       - name: Deploy
         uses: JamesIves/github-pages-deploy-action@4.1.3
         with:
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: pnpm/action-setup@v2
+      - uses: pnpm/action-setup@v4
       - name: Root Setup
         run: |
           pnpm install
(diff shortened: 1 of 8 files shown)


samchon (commented):
I think `nestia` needs matrix strategy of NodeJS versions. Others are good to see.

samchon (changes requested):
Also, ignoring `pnpm-lock.yaml` file is inteded.

As this is a library repository, I've always test the `^` version setup startegy.

ryoppippi:
OK, then use another pr!

ryoppippi:
I'll add matrix later

ryoppippi:
So managing node version by pnpm is not good for this case?

samchon (changes requested):
As `nestia` utilizes many NodeJS built-in functions, I think NodeJS version matrix is valid for this repo.

Also, do not change too much style please. Ignoring `pnpm-lock.yaml` file is required for dependencies' patch and minor updates.

.github/workflows/website.yml
samchon, line 1: Also, website must use `npm` at now.

This may be caused by my unskilled monorepo composition, anyway, if try to use `pnpm install` command in the `website` directory, it does not install of `website`, but of other modules enrolled in the `pnpm-workspace.yaml` file.

By the way, if add the `website` directory to the `pnpm-workspace.yaml` file is impossible. It is because the `website` module requires other module's complete compilation and testing.

pnpm-lock.yaml
samchon, line 1: This must be removed with below commented reason.

.gitignore
samchon, line 1: Also change please

.github/workflows/website.yml
```diff
@@ -12,19 +12,19 @@ jobs:
...
+
+      - uses: pnpm/action-setup@v4
       - name: Root Setup
-        run: pnpm install && npm run build
+        run: |
+          pnpm install
+          pnpm run build
       - name: Build
         working-directory: website 
-        run: npm install && npm run build
```
samchon, line 25: Also, website must use `npm` at now.

This may be caused by my unskilled monorepo composition, anyway, if try to use `pnpm install` command in the `website` directory, it does not install of `website`, but of other modules enrolled in the `pnpm-workspace.yaml` file.

By the way, if add the `website` directory to the `pnpm-workspace.yaml` file is impossible. It is because the `website` module requires other module's complete compilation and testing.

samchon (changes requested):
Thanks for your endless contribution.

There may be a better and more superior solution, or I may have made a wrong choice.

However, the changes must be made at a level that I can keep going the development.

ryoppippi:
I'll create another PR based on this!

PR #1408, opened 2025-12-11
Update TypedBody.mdx

niraj-khatiwada:
Switch `IBbsArticle.ts`, `IAttachmentFile.ts` tab position to show right interface example in the [documentation](https://nestia.io/docs/core/TypedBody/#how-to-use).
image


samchon (approved):
Oops, they were reversed. Thanks for fixing.

PR #1454, opened 2026-04-18
feat(core, sdk): add Model Context Protocol (MCP) support

### LINKED ISSUE #1350: Support MCP route functions from `@nestia/core`
Suggest proper decorator functions for MCP routing, and implement them.

wildduck2:
This pull request adds Model Context Protocol (MCP) support to nestia. It closes #1350 and supersedes the earlier draft at #1351.

The change is split into three commits.

The first commit adds `@McpRoute` to `@nestia/core`. The decorator marks a controller method as an MCP tool. At bootstrap, `McpAdaptor.upgrade(app)` walks the Nest container, collects every method carrying the decorator, and serves them over a Streamable HTTP endpoint provided by the official `@modelcontextprotocol/sdk` server. The decorator is backed by typia at compile time through a new `McpRouteProgrammer` that emits JSON Schema 2020-12 for the parameter type, and a new `McpRouteTransformer` that wires the programmer into the `ts-patch` pipeline. Arguments are validated at runtime via typia's `assert` path, reusing the same `validate_request_body` plumbing as `@TypedBody`. Error mapping follows the MCP specification: invalid arguments surface as JSON-RPC `-32602`, unknown tool names as `-32601`, `HttpException` thrown by a handler becomes a tool result with `isError: true`, and any other throw becomes `-32603`.

The second commit extends `@nestia/sdk` so that `nestia sdk`, `nestia swagger`, and `nestia all` understand the new protocol. New structures `IReflectMcpOperation`, `IReflectMcpOperationParameter`, and `ITypedMcpRoute` describe the reflected operation and the final typed route. New analyzers `ReflectMcpOperationAnalyzer` and `TypedMcpRouteAnalyzer` read the `"nestia/McpRoute"` metadata and turn it into `ITypedMcpRoute` with a stable accessor of the form `["mcp", tool]`. Existing dispatch points in `ReflectControllerAnalyzer`, `NestiaSdkApplication`, `NestiaSwaggerComposer`, `AccessorAnalyzer`, `SdkFileProgrammer`, and `SdkRouteDirectory` are updated to handle the third protocol branch. The swagger composer filters MCP routes out of the HTTP document, matching the policy already used for WebSocket routes. A new `SdkMcpRouteProgrammer` emits, for each tool, an async wrapper that takes an MCP `Client` plus typed arguments, calls the tool, narrows the response structurally, and returns `Primitive<T>`. The generated code avoids casts on the response; the only remaining type assertion is the `arguments` bridge required by the MCP SDK index-signature type.

The third commit adds `tests/test-sdk/features/mcp`, mirroring `tests/test-sdk/features/websocket`. `Backend.ts` stands up a minimal NestJS application and calls `McpAdaptor.upgrade` on port `37000`. Two controllers exercise flat and nested input shapes, and `divide` throws `BadRequestException` on zero so the `isError` path is covered. Twelve tests run in two layers. Six of them call the MCP server directly through `@modelcontextprotocol/sdk` and cover the wire protocol: `initialize`, `tools/list`, `tools/call`, and the three error paths for unknown tool, invalid arguments, and domain errors. The other six use the generated `api.functional.mcp.*` wrappers and cover `add`, `subtract`, `divide`, `get_weather` with nested input, `METADATA` literal assertions, and the SDK wrapper's thrown error on `isError`.

## Usage

```ts
import core from "@nestia/core";

interface ICalcInput { a: number; b: number; }
interface ICalcResult { result: number; }

@Controller()
export class CalculatorController {
  @core.McpRoute({
    name: "add",
    description: "Return the sum of two numbers.",
  })
  public async add(
    @core.McpRoute.Params() params: ICalcInput,
  ): Promise<ICalcResult> {
    return { result: params.a + params.b };
  }
}
```

```ts
import core from "@nestia/core";
import { NestFactory } from "@nestjs/core";

const app = await NestFactory.create(AppModule);
await core.McpAdaptor.upgrade(app, { path: "/mcp" });
await app.listen(3000);
```

Generated SDK in `src/api/functional/mcp/index.ts`:

```ts
export async function add(
  client: Client,
  args: add.Input,
): Promise<add.Output> { ... }
export namespace add {
  export type Input = ICalcInput;
  export type Output = Primitive<ICalcResult>;
  export const METADATA = {
    protocol: "mcp",
    tool: "add",
    title: null,
    description: "Return the sum of two numbers.",
  } as const;
}
```

## Verification

- `pnpm --filter @nestia/core build` passes.
- `pnpm --filter @nestia/sdk build` passes.
- `cd tests/test-sdk/features/mcp && npx nestia all` succeeds, producing sdk, swagger, and e2e outputs.
- `cd tests/test-sdk/features/mcp && npx ts-node src/test` reports twelve passing tests.
- `cd tests/test-sdk && node start.js` runs the MCP feature end to end.

Closes #1350.
Supersedes #1351.


samchon (changes requested):
About description, it would better to parsing JSDoc comments instead of let user to fill the `description` property.

wildduck2:
Thanks for the review @samchon. You were right, the `description` field on the decorator config was redundant when the method already carries a JSDoc block. I pushed a follow-up commit that parses the JSDoc at transform time and injects `description` (from the main comment) and `title` (from an optional `@title` tag) into the `@McpRoute(...)` config object when the user did not set them explicitly. Explicit values still win, so the override path is preserved.


After the change, controllers look like:

```ts
@Controller()
export class CalculatorController {
  /** Return the sum of two numbers. */
  @core.McpRoute({ name: "add" })
  public async add(
    @core.McpRoute.Params() params: ICalcInput,
  ): Promise<ICalcResult> {
    return { result: params.a + params.b };
  }
}
```

All twelve MCP feature tests still pass (`npx ts-node src/test` reports `Success`) and the generated SDK at `src/api/functional/mcp/index.ts` keeps the description both on the function's emitted JSDoc and in the `METADATA` constant.

samchon (changes requested):
## Limitation following the MCP schema protocol

At first, `@McpRoute.Params()` must be allowed only one time in a function. This is blocked in SDK generation, but not blocked consistently in the `@nestia/core` MCP path itself.

Also, parameters must be an object type without dynamic properties. You need to do the same thing as `TypedBodyProgrammer.generate()` by utilizing `LlmSchemaProgrammer.validate()`.

Such limitation must also be applied to the return type. In MCP, the return type must be `void` or an object type without dynamic properties. A union type of both must not be allowed. You can follow the same logic as `TypedRouteProgrammer.generate()`.

Please also add compile-fail feature tests for these constraints. The existing `tests/test-sdk` suite already uses `*-error-*` features for this style of validation, so MCP should add analogous cases instead of relying only on runtime happy-path tests.

- `mcp-error-multiple-params`
- `mcp-error-param-non-object`
- `mcp-error-param-dynamic-properties`
- `mcp-error-return-non-object`
- `mcp-error-return-dynamic-properties`
- `mcp-error-return-union-void-object`

## Interface of `@McpRoute()`

Public API of `@McpRoute()` should accept only parameter `string`.

If you wanna provide `title`, let user write a JSDoc tag like `@title something`.

```ts
/**
 * Description...
 */
export function McpRoute(name: string): MethodDecorator;

/**
 * @internal
 */
export function McpRoute(config: McpRoute.IConfig): MethodDecorator;

export namespace McpRoute {
  /**
   * @internal
   */
  export interface IConfig {}
}
```

wildduck2:
> ## Limitation following the MCP schema protocol
> At first, `@McpRoute.Params()` must be allowed only one time in a function. This is blocked in SDK generation, but not blocked consistently in the `@nestia/core` MCP path itself.
> 
> Also, parameters must be an object type without dynamic properties. You need to do the same thing as `TypedBodyProgrammer.generate()` by utilizing `LlmSchemaProgrammer.validate()`.
> 
> Such limitation must also be applied to the return type. In MCP, the return type must be `void` or an object type without dynamic properties. A union type of both must not be allowed. You can follow the same logic as `TypedRouteProgrammer.generate()`.
> 
> Please also add compile-fail feature tests for these constraints. The existing `tests/test-sdk` suite already uses `*-error-*` features for this style of validation, so MCP should add analogous cases instead of relying only on runtime happy-path tests.
> 
> * `mcp-error-multiple-params`
> * `mcp-error-param-non-object`
> * `mcp-error-param-dynamic-properties`
> * `mcp-error-return-non-object`
> * `mcp-error-return-dynamic-properties`
> * `mcp-error-return-union-void-object`
> 
> ## Interface of `@McpRoute()`
> Public API of `@McpRoute()` should accept only parameter `string`.
> 
> If you wanna provide `title`, let user write a JSDoc tag like `@title something`.
> 
> ```ts
> /**
>  * Description...
>  */
> export function McpRoute(name: string): MethodDecorator;
> 
> /**
>  * @internal
>  */
> export function McpRoute(config: McpRoute.IConfig): MethodDecorator;
> 
> export namespace McpRoute {
>   /**
>    * @internal
>    */
>   export interface IConfig {}
> }
> ```

I will spare some time to do this, soon

samchon:
MCP route branch conflict resolution and several review fixes were pushed in `421ccd99845aa02374ccea9b0975326f27fecaf2`.

Implemented:
- Merged current `samchon/nestia` master into the branch.
- Generated MCP SDK imports now alias MCP SDK symbols (`Client as McpClient`, `CallToolResult as McpCallToolResult`) so user DTO names can still be `Client` / `CallToolResult`.
- `@McpRoute.Params()` is now required to be exactly one method parameter. Missing decorator, no params, extra params, multiple params, non-object params, and dynamic-key params are compile-fail fixtures.
- `@nestia/core` keeps `@modelcontextprotocol/sdk` only as a dev dependency, not a peer dependency. Runtime `McpAdaptor.upgrade()` lazy-loads the MCP SDK and reports a clear error only when MCP is actually enabled without the SDK installed.
- SDK distribution package generation adds `@modelcontextprotocol/sdk` to generated package `dependencies` only when the analyzed app contains MCP routes; the static distribution template is not polluted.
- Generated `@controller` comments now use the actual controller method name rather than the SDK accessor name.
- Windows `test-sdk` runner no longer breaks on `ts-node -O` shell quoting.

Questions / deferred review items:
- `sessioned: true`: the adaptor still creates a fresh `McpServer`/transport per request. If session mode must be supported, we likely need persistent transport/session storage across requests. Is that intended for this PR, or should the option be removed/deferred until stateful transport handling is designed?
- `void` return tools: compile validation currently allows `void`, but the adaptor/generated SDK path is fundamentally JSON text oriented. Should void MCP tools be represented as an empty object/structured sentinel, or should `void` be rejected for MCP routes for now?
- Duplicate MCP tool names: current registration/call lookup would allow duplicates and effectively call the first matching tool. Should we make duplicate `McpRoute({ name })` a compile/SDK analysis error?

Verified locally:
- `pnpm --filter @nestia/core build`
- `pnpm --dir tests/test-sdk start -- --only mcp`

samchon:
Follow-up implementation report. I pushed the requested fixes and the additional self-review fixes in `421ccd99`, `54e53884`, `caf32b85`, and `151aca33`.

## 1. Master conflict resolution

I merged the current `samchon/nestia` `master` into this branch and adjusted the MCP implementation to match the current SDK/distribution flow.

I intentionally did not modify the static `distribute/package.json` template. Instead, SDK distribution composition now installs `@modelcontextprotocol/sdk` only when the analyzed application contains MCP routes, using `app.routes.some((r) => r.protocol === "mcp")`. Generated SDK packages without MCP routes do not receive the MCP dependency.

## 2. MCP SDK dependency and alias imports

`@nestia/core` and `@nestia/sdk` keep `@modelcontextprotocol/sdk` as a development dependency because this repository needs it for type-checking and tests. The published runtime path does not force ordinary `@nestia/core` users to install it: `McpAdaptor.upgrade()` loads the MCP SDK lazily and throws a clear error only when MCP is actually enabled without the dependency installed.

Generated MCP SDK clients alias MCP SDK types so user DTO names can still be `Client` or `CallToolResult`:

```ts
import type { Client as McpClient } from "@modelcontextprotocol/sdk/client/index.js";
import type { CallToolResult as McpCallToolResult } from "@modelcontextprotocol/sdk/types.js";
```

This is covered by `tests/test-sdk/features/mcp/src/controllers/ImportAliasController.ts` and `test_api_mcp_alias_imports.ts`.

## 3. `sessioned` was removed/deferred

The original `sessioned` option implied stateful MCP Streamable HTTP support, but the adaptor did not actually implement a safe stateful server. Proper stateful support needs:

- a transport/server registry keyed by `Mcp-Session-Id`;
- session lookup after `initialize`;
- TTL and cleanup policy;
- resource cleanup on close/error;
- deployment guidance for sticky sessions or an external session store.

Because this PR does not implement that machinery, the type-level option was removed and runtime now rejects a `sessioned` key explicitly. Documentation now describes the current scope as a stateless Streamable HTTP tools endpoint.

## 4. `@McpRoute.Params()` contract

MCP tools now require exactly one method parameter, and that parameter must be decorated with `@McpRoute.Params()`.

The parameter type must be a static object type without dynamic keys. `Record<string, T>` and index signatures are rejected because they do not produce a stable MCP tool argument schema or generated SDK contract.

Additional compile-fail fixtures were added:

- `mcp-error-duplicate-tool-name`: duplicate MCP tool names are rejected.
- `mcp-error-duplicate-accessor`: distinct tool names that collapse to the same generated SDK accessor are rejected, for example `same-name` and `same_name` both becoming `api.functional.mcp.same_name`.

The second case came from self-review. MCP runtime dispatch could still distinguish the raw tool names, but the generated TypeScript SDK would emit duplicate function names. The generator now reports this during validation with the offending `@McpRoute(...)` location.

## 5. `void` return handling

When an MCP handler returns `void`/`undefined`, the server returns `{ content: [] }`. It does not synthesize JSON text content, because the declared contract is that the tool returns no payload.

Generated SDK wrappers also treat this as a first-class `void` contract: they return `Promise<void>` and emit `return;` instead of reading/parsing `result.content[0]`.

This is covered at both layers:

- wire test: `client.callTool({ name: "notify" })` returns `content: []`;
- generated SDK test: `api.functional.mcp.notify(...)` returns `undefined`.

## 6. Public API examples and fixtures

Self-review found that the website docs and some test fixtures were still using the internal transformer-emitted object form, `@McpRoute({ name: "..." })`.

The public authoring API is `@McpRoute("tool_name")`, so docs and fixtures were changed to that form.

That change also exposed an existing transformer fragility: `hasField()` called `getText()` on synthetic object-literal property names. Synthetic nodes may not have source text. The transformer now reads property names through `Identifier.escapedText` or string literal `.text`, making the string-overload rewrite path stable.

## 7. Website documentation

Added `website/src/content/docs/core/McpRoute.mdx` and exposed it in `_meta.ts`. The page documents:

- `@McpRoute()` and `@McpRoute.Params()`;
- `McpAdaptor.upgrade()`;
- the required application-level `@modelcontextprotocol/sdk` install when MCP is enabled;
- parameter and return type restrictions;
- generated SDK usage;
- stateless endpoint scope;
- unique tool-name requirements.

## 8. Deferred scope

Making Nestia a full stateful MCP server platform is intentionally out of scope for this PR. MCP resources, prompts, sampling, elicitation, and stateful session transport need a separate design.

This PR implements the narrower and currently defensible scope: expose NestJS controller methods as stateless MCP tools through `@McpRoute`.

## Verification

Local verification:

```bash
pnpm --filter @nestia/core build
pnpm --filter @nestia/sdk build
pnpm --dir tests/test-sdk start -- --only mcp
git diff --check
```

CI was then approved for the fork PR and completed successfully. The PR has been merged.

samchon (approved):
Requested MCP follow-up changes have been addressed: stateless contract, strict Params constraints, SDK alias imports, void returns, duplicate tool/accessor validation, conditional distribution dependency, and website documentation. Local focused MCP tests passed; waiting for the remaining full CI test job before merge.

samchon:
Follow-up note for the `next` branch: this MCP route feature should be implemented once more against the TypeScript-Go / `ttsc` migration pipeline instead of being treated as fully done by this merge alone.

I opened #1478 for that work and left a detailed next-session prompt there. The intent is to port the same public contract (`@McpRoute(tool_name)`, exactly one `@McpRoute.Params()` object parameter, stateless `McpAdaptor`, aliased MCP SDK imports, `void` returns, duplicate tool/accessor validation, and conditional generated SDK dependency injection) into the `next` branch architecture after inspecting its Go-backed compiler/transformer flow.

PR #1479, opened 2026-06-11
fix(sdk): validate keyword body assertions

### LINKED ISSUE #1410: [BUG] SDK compile error when option the nestia config option assert is set to tur (`assert: true`)
### Summary
If you set the **assert**-option to **true** in the nestia config, there are compile errors in the generated SDK files:  **dto not found** in the auto generated **functional** folder in line `typia.assert<typeof dto>(dto);`, because the method header does only contain `connection` and `props` and not an `dto`.

```typescript
  input: async () => {
    const app = await NestFactory.create(ApplicationModule, new FastifyAdapter());
    app.setGlobalPrefix('api');
    return app;
  },

  // SDK output folder
  output: 'sdk/api',
  primitive: false,
  clone: true,
  assert: true,
  json: true,
  keyword: true,

  distribute: 'packages/api',
};
```

Therefore, the command `npm run compile` fails in the auto-generated API package..

### Code occuring the bug
The controller with the corresponding endpoint:
```typescript
import { TypedBody, TypedRoute } from '@nestia/core';
import { Controller } from '@nestjs/common';
import { tags } from 'typia';

export class ResponseDto {
  status: number;
}

export class CreateDto {
  name: string;
}

@Controller('example')
export class ExampleController {
  @TypedRoute.Post()
  async create(@TypedBody() dto: CreateDto): Promise<ResponseDto> {
    return { status: 201 };
  }
}
```

The auto generated file form nestia containing the error `sdk/api/functional/api/example`:
```typescript
/**
 * @packageDocumentation
 * @module api.functional.api.example
 * @nestia Generated by Nestia - https://github.com/samchon/nestia
 */
//================================================================
import type { IConnection } from "@nestia/fetcher";
import { NestiaSimulator } from "@nestia/fetcher/lib/NestiaSimulator";
import { PlainFetcher } from "@nestia/fetcher/lib/PlainFetcher";
import typia from "typia";

import type { CreateDto } from "../../../structures/CreateDto";
import type { ResponseDto } from "../../../structures/ResponseDto";

/**
 * @controller ExampleController.create
 * @path POST /api/example
 * @accessor api.functional.api.example.create
 * @nestia Generated by Nestia - https://github.com/samchon/nestia
 */
export async function create(
  connection: IConnection,
  props: create.Props,
): Promise<create.Output> {
  typia.assert<typeof dto>(dto); // <-- error dto not found
  return true === connection.simulate
    ? create.simulate(connection, props)
    : PlainFetcher.fetch(
        {
          ...connection,
          headers: {
            ...connection.headers,
            "Content-Type": "application/json",
          },
        },
        {
          ...create.METADATA,
          template: create.METADATA.path,
          path: create.path(),
        },
        props.dto,
        create.stringify,
      );
}
export namespace create {
  export type Props = {
    dto: Body;
  };
  export type Body = CreateDto;
  export type Output = ResponseDto;

  export const METADATA = {
    method: "POST",
    path: "/api/example",
    request: {
      type: "application/json",
      encrypted: false,
    },
    response: {
      type: "application/json",
      encrypted: false,
    },
    status: 201,
  } as const;

  export const path = () => "/api/example";
  export const random = (): ResponseDto => typia.random<ResponseDto>();
  export const simulate = (connection: IConnection, props: Props): Output => {
    const assert = NestiaSimulator.assert({
      method: METADATA.method,
      host: connection.host,
      path: path(),
      contentType: "application/json",
    });
    assert.body(() => typia.assert(props.dto));
    return random();
  };
  export const stringify = (input: Body) => typia.json.stringify(input);
}
```

samchon:
## Summary

Fixes #1410.

- Reuses the generated SDK parameter accessor when emitting config-level `typia.assert` calls.
- Makes keyword-mode SDK clients assert `props.dto` instead of an out-of-scope `dto` identifier.
- Extends the `keyword` SDK fixture with `assert: true` so CI covers the combined mode.

## Verification

- `pnpm --filter @nestia/core run build`
- Generator smoke script confirmed `typia.assert<typeof props.dto>(props.dto)` output.

## Not Run

- `tests/test-sdk` keyword fixture generation on this Windows checkout; blocked before SDK output by existing native plugin temp filename normalization diagnostics (`fileName should be normalized and absolute` for files under `%TEMP%`).

### DIFF
--- a/packages/sdk/src/generates/internal/SdkHttpFunctionProgrammer.ts
+++ b/packages/sdk/src/generates/internal/SdkHttpFunctionProgrammer.ts
@@ -1,6 +1,12 @@
-import { Node, NodeFlags, SyntaxKind, TypeScriptFactory } from "@nestia/factory";
+import {
+  Node,
+  NodeFlags,
+  SyntaxKind,
+  TypeScriptFactory,
+} from "@nestia/factory";
 import { IdentifierFactory, TypeFactory } from "@nestia/factory";
 
+import { sizeOf } from "../../internal/legacy";
 import { INestiaProject } from "../../structures/INestiaProject";
 import { ITypedHttpRoute } from "../../structures/ITypedHttpRoute";
 import { StringUtil } from "../../utils/StringUtil";
@@ -8,7 +14,6 @@ import { ImportDictionary } from "./ImportDictionary";
 import { SdkAliasCollection } from "./SdkAliasCollection";
 import { SdkHttpParameterProgrammer } from "./SdkHttpParameterProgrammer";
 import { SdkImportWizard } from "./SdkImportWizard";
-import { sizeOf } from "../../internal/legacy";
 
 export namespace SdkHttpFunctionProgrammer {
   export const write =
@@ -202,12 +207,8 @@ export namespace SdkHttpFunctionProgrammer {
                     ),
                     "assert",
                   ),
-                  [
-                    TypeScriptFactory.createTypeQueryNode(
-                      TypeScriptFactory.createIdentifier(p.name),
-                    ),
-                  ],
-                  [TypeScriptFactory.createIdentifier(p.name)],
+                  [TypeScriptFactory.createTypeQueryNode(access(p.name))],
+                  [access(p.name)],
                 ),
               ),
             )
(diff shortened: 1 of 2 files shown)


samchon:
Post-CI self-review completed.

Pass 1/3: checked the generator diff against the reported failure. `typia.assert` now uses the same parameter accessor as generated fetch arguments, so keyword mode emits `props.dto` and positional mode still emits `dto`.

Pass 2/3: checked regression coverage. Adding `assert: true` to the existing `keyword` SDK fixture covers the combined mode that previously emitted an out-of-scope identifier.

Pass 3/3: checked blast radius and CI. The change is limited to config-level SDK assertion operands plus fixture config, and all PR checks are passing.

No follow-up changes needed.

PR #1480, opened 2026-06-11
fix(fetcher): preserve Set-Cookie response headers

### LINKED ISSUE #863: `response_headers_to_object` can be improved
I think `response_headers_to_object` is not good enough. Actually it is quite uncomfortable when multiple cookies with options are set by `Set-Cookie` header.

https://github.com/samchon/nestia/blob/fbe753e45807c3a92d54957c4643bb8423c89352/packages/fetcher/src/internal/FetcherBase.ts#L256

## Problems
When multiple cookies are set without any option, then it's fine.

For example,
```ts
res.setCookie('cookie1', 'qwe123');
res.setCookie('cookie2', 'asd123');
```
this will be converted to:
```
'set-cookie': [
    'cookie1=qwe123',
    'cookie2=asd123'
]
```

However, when cookie security options are set along like this:
```ts
res.setCookie('cookie1', 'qwe123', {
    httpOnly: true,
    secure: true,
    sameSite: 'lax',
    path: '/',
    expires: new Date(),
});
res.setCookie('cookie2', 'asd123', {
    httpOnly: true,
    secure: true,
    sameSite: 'lax',
    path: '/',
    expires: new Date(),
});
```
now things get messy
```
'set-cookie': [
    'cookie1=qwe123',
    'HttpOnly',
    'Secure',
    'SameSite=Lax',
    'Path=/',
    'Expires=Fri, 05 Apr 2024 08:31:47 GMT',
    'cookie2=asd123',
    'HttpOnly',
    'Secure',
    'SameSite=Lax',
    'Path=/',
    'Expires=Fri, 05 Apr 2024 08:31:47 GMT',
]
```
This is not comfortable as it is hard to tell which cookie each elements belong to.

## Suggestions
Fortunately, `response_headers_to_object` is parsing multiple cookies one by one
```
key: set-cookie
value: cookie1=qwe123; Path=/; Expires=Fri, 05 Apr 2024 12:31:46 GMT; HttpOnly; Secure; SameSite=Lax
key: set-cookie
value: cookie2=asd123; Path=/; Expires=Fri, 05 Apr 2024 12:31:46 GMT; HttpOnly; Secure; SameSite=Lax
```

So, I would suggest either
### 1. pushing cookie strings
```ts
const response_headers_to_object = (
  headers: Headers,
): Record<string, string | string[]> => {
  const output: Record<string, string | string[]> = {};
  headers.forEach((value, key) => {
    if (key === "set-cookie") {
      output[key] ??= [];
      (output[key] as string[]).push(value);
    } else output[key] = value;
  });
  return output;
};
```
```ts
'set-cookie': [
    'cookie1=qwe123; Path=/; Expires=Fri, 05 Apr 2024 12:31:46 GMT; HttpOnly; Secure; SameSite=Lax',
    'cookie2=asd123; Path=/; Expires=Fri, 05 Apr 2024 12:31:46 GMT; HttpOnly; Secure; SameSite=Lax'
]
```
So, let people handle their cookie themselves,
or

### 2. construct as object
```ts
const response_headers_to_object = (
  headers: Headers,
): Record<string, string | string[] | Record<string, any>> => {
  const output: Record<string, string | string[] | Record<string, any>> = {};
  headers.forEach((value, key) => {
    if (key === "set-cookie") {
      output[key] ??= {};
      const parsedCookie = parseCookies([value]);
      Object.assign(output[key], parsedCookie);
    } else output[key] = value;
  });
  return output;
};

type CookieOptions = CookieOption['options'] & { value: string };
export const parseCookies = (cookies: string[]) => {
  return cookies.reduce(
    (acc, cookie) => {
      const parts = cookie.split(';');
      const keyValue = parts.shift();
      if (keyValue === undefined) return acc;

      const [key, value] = keyValue.split('=');
      acc[key] = parts.reduce(
        (acc, part) => {
          const splits = part.trim().split('=');
          const optionKey = splits[0].replace(/[A-Z]/, (match) => {
            return match.toLowerCase();
          }) as keyof CookieOption['options'];
          const optionValue = splits[1];
          switch (optionKey) {
            case 'expires':
              acc.expires = new Date(optionValue);
              break;
            case 'httpOnly':
            case 'secure':
              acc[optionKey] = true;
              break;
            default:
              acc[optionKey] = optionValue as any;
          }
          return acc;
        },
        { value: value.trim() } as CookieOptions,
      );

      return acc;
    },
    {} as Record<string, CookieOptions>,
  );
};
```
```
'set-cookie': {
    cookie1: {
      value: 'qwe123',
      path: '/',
      expires: 2024-04-05T09:14:04.000Z,
      httpOnly: true,
      sameSite: 'Lax'
    },
    cookie2: {
      value: 'asd123',
      path: '/',
      expires: 2024-04-05T09:14:04.000Z,
      httpOnly: true,
      sameSite: 'Lax'
    }
  },
```
and possibly better if you could provide type-safe getter.

samchon:
## Summary

Fixes #863.

- Preserves each `Set-Cookie` response header as a complete cookie string.
- Uses `Headers.getSetCookie()` when the runtime exposes it, with a lossless `forEach()` fallback.
- Adds an e2e regression test covering two cookies with semicolon-delimited attributes.

## Verification

- `pnpm --filter @nestia/fetcher run build`
- Direct `ttsx` project check for `tests/test-e2e`
- `PlainFetcher.propagate` smoke test with two `Set-Cookie` headers

## Not Run

- `pnpm --filter ./tests/test-e2e start` through the package script on Windows; the script overwrites `TTSC_GO_BINARY` with a non-absolute `go` path, which this checkout cannot resolve.

### DIFF
--- a/packages/fetcher/src/IPropagation.ts
+++ b/packages/fetcher/src/IPropagation.ts
@@ -47,7 +47,7 @@ export type IPropagation<
         StatusMap[P]
       >;
     }[keyof StatusMap]
-  | IPropagation.IBranch<false, unknown, unknown>;
+  | IPropagation.IBranch<false, number, unknown>;
 export namespace IPropagation {
   /**
    * Type of configurable status codes.
--- a/packages/fetcher/src/internal/FetcherBase.ts
+++ b/packages/fetcher/src/internal/FetcherBase.ts
@@ -223,12 +223,17 @@ const response_headers_to_object = (
   headers: Headers,
 ): Record<string, string | string[]> => {
   const output: Record<string, string | string[]> = {};
+  const cookieHeaders: string[] | undefined =
+    typeof (headers as any).getSetCookie === "function"
+      ? (headers as any).getSetCookie()
+      : undefined;
+  if (cookieHeaders?.length) output["set-cookie"] = cookieHeaders;
+
   headers.forEach((value, key) => {
-    if (key === "set-cookie") {
+    if (key.toLowerCase() === "set-cookie") {
+      if (cookieHeaders?.length) return;
       output[key] ??= [];
-      (output[key] as string[]).push(
-        ...value.split(";").map((str) => str.trim()),
-      );
+      (output[key] as string[]).push(value);
     } else output[key] = value;
   });
   return output;
--- a/tests/test-e2e/src/features/test_fetcher_propagation_unknown_status.ts
+++ b/tests/test-e2e/src/features/test_fetcher_propagation_unknown_status.ts
@@ -1,0 +1,26 @@
+import { IPropagation } from "@nestia/fetcher";
+import typia from "typia";
+
+/**
+ * Verifies propagation accepts undeclared failure status codes.
+ *
+ * Locks the fallback branch of `IPropagation`. Generated SDK e2e tests assert
+ * propagated outputs with typia, and HTTP servers can still return statuses
+ * that are not listed in the route's explicit exception map, such as payload
+ * size errors from the platform.
+ *
+ * 1. Build an `IPropagation` value with a declared 200 success branch.
+ * 2. Fill it with an undeclared 413 failure response.
+ * 3. Assert typia accepts the fallback failure branch.
+ */
+export async function test_fetcher_propagation_unknown_status(): Promise<void> {
+  typia.assert<IPropagation<{ 200: string }, 200>>({
+    success: false,
+    status: 413,
+    headers: {},
+    data: {
+      statusCode: 413,
+      message: "Payload Too Large",
+    },
+  });
+}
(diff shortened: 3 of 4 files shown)


samchon:
CI follow-up pushed.

The previous `sdk` failure was in `clone-and-propagate-and-keyword`: typia rejected a propagated HTTP 413 response because `IPropagation`'s fallback failure branch used `unknown` as the status type, which reduced `status` to `never`. I changed the fallback branch to accept numeric statuses and added a focused e2e type-regression test for undeclared propagated failures.

samchon:
Self-review after green CI (3 passes):

1. Header normalization: verified Set-Cookie no longer splits on semicolons, uses getSetCookie() when the runtime exposes it, and falls back without corrupting cookie attributes.
2. Propagation typing: verified undeclared failure statuses remain in the false branch with numeric status, so generated SDK e2e typia assertions accept platform errors like 413 without widening success branches.
3. Test coverage: verified focused e2e tests cover full Set-Cookie strings and undeclared propagated failure status. CI is green across sdk, e2e, go, migrate, benchmark, transform-options, deploy, Ubuntu, spell-check, and Socket checks.

PR #1484, opened 2026-06-11
fix(sdk): honor global prefix exclusions

### LINKED ISSUE #746: "exclude" option not considered while using app.setGlobalPrefix(..) in nestia.config.ts
### Summary
When calling the `app.setGlobalPrefix(...)` function in `nestia.config.ts`, it appears that the paths mentioned in the "exclude" option are also getting the global prefix (see example below). The correct behavior is that such paths should be excluded from the global prefix. When the SDK generates the fetch function for such routes, it should not apply the global prefix to their paths.

  - **SDK Version**: "@nestia/sdk": "^2.4.5",

  - **Expected behavior**: The sdk should not apply the global prefix to the paths included in the "exclude" option. 

  - **Actual behavior**: The paths included in the "exclude" option should be ignored while setting the global prefix, but they are not being excluded in the generated sdk route.

### Code where the bug is occuring:
````ts
import { INestiaConfig } from '@nestia/sdk';
import { VERSION_NEUTRAL, VersioningType } from '@nestjs/common';
import { NestFactory } from '@nestjs/core';
import { initializeApp } from 'firebase-admin/app';
import { AppModule } from './src/app/app.module';

  input: async () => {
    initializeApp();

    const app = await NestFactory.create(AppModule, {
      bodyParser: false,
    });

    app.setGlobalPrefix('api', { exclude: ['/_ah/warmup'] });  // 👈 here 

    app.enableVersioning({
      type: VersioningType.URI,
      prefix: 'v',
      defaultVersion: VERSION_NEUTRAL,
    });

    return app;
  },
  output: 'libs/shared/client/api/src',
  clone: true,
  primitive: true,
};
export default NESTIA_CONFIG;
````

samchon:
Closes #746.

## Summary
- apply Nest's `setGlobalPrefix(..., { exclude })` metadata while composing SDK and Swagger paths
- compare exclusions against the local unprefixed route path, matching Nest's URI-versioning behavior
- add an `app-globalPrefix` regression route for `/_ah/warmup` and assert SDK + Swagger stay unprefixed

## Tests
- `pnpm --filter @nestia/sdk run build` with absolute `TTSC_GO_BINARY`
- Node smoke test for excluded, prefixed, and versioned `PathAnalyzer.joinWithGlobalPrefix` cases

## Not Run
- `node tests/test-sdk/start.js --only app-globalPrefix --concurrency 1` could not complete locally on Windows because the native transform rejected generated temp filenames before fixture generation (`fileName should be normalized and absolute`).

### DIFF
--- a/packages/sdk/src/NestiaSdkApplication.ts
+++ b/packages/sdk/src/NestiaSdkApplication.ts
@@ -1,4 +1,3 @@
-import { IMetadataDictionary } from "./internal/legacy";
 import fs from "fs";
 import path from "path";
 import { HashSet, Pair, TreeMap } from "tstl";
@@ -13,13 +12,14 @@ import { TypedWebSocketRouteAnalyzer } from "./analyses/TypedWebSocketRouteAnaly
 import { E2eGenerator } from "./generates/E2eGenerator";
 import { SdkGenerator } from "./generates/SdkGenerator";
 import { SwaggerGenerator } from "./generates/SwaggerGenerator";
+import { IMetadataDictionary } from "./internal/legacy";
 import { INestiaProject } from "./structures/INestiaProject";
+import { IOperationMetadata } from "./structures/IOperationMetadata";
 import { IReflectController } from "./structures/IReflectController";
 import { IReflectOperationError } from "./structures/IReflectOperationError";
 import { ITypedApplication } from "./structures/ITypedApplication";
 import { ITypedHttpRoute } from "./structures/ITypedHttpRoute";
 import { ITypedWebSocketRoute } from "./structures/ITypedWebSocketRoute";
-import { IOperationMetadata } from "./structures/IOperationMetadata";
 import { StringUtil } from "./utils/StringUtil";
 import { VersioningStrategy } from "./utils/VersioningStrategy";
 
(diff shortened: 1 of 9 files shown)


samchon:
Post-CI self-review completed.

Pass 1/3: Checked route composition semantics. The new helper compares the unprefixed local route against Nest's exclude metadata and applies the global prefix only when the route is not excluded.

Pass 2/3: Checked SDK and Swagger coverage. Both `NestiaSdkApplication` and `NestiaSwaggerComposer` use the same helper, and the `app-globalPrefix` fixture locks SDK + Swagger output for `/_ah/warmup`.

Pass 3/3: Checked verification and CI. All checks passed (`sdk`, `e2e`, `go`, `migrate`, `transform-options`, `benchmark`, `Ubuntu`, `deploy`, `spell-check`, Socket). No follow-up changes needed.

samchon:
Self-review after rebase and green CI (3 passes):

1. Path composition: verified global prefix exclusion compares the unversioned local route path against Nest's exclude metadata, then applies prefix only to non-excluded routes.
2. Protocol/method handling: verified Swagger composition now filters to HTTP operations before reading method data, while SDK path composition keeps WebSocket routes unaffected by passing a non-HTTP method.
3. Coverage: verified the app-globalPrefix fixture calls the excluded warmup route and checks Swagger exposes /_ah/warmup without /x/_ah/warmup. CI is green across sdk, e2e, go, migrate, benchmark, transform-options, deploy, Ubuntu, spell-check, and Socket checks.

PR #1486, opened 2026-06-11
test(core): cover repeated TypedQuery arrays

### LINKED ISSUE #1404: TypedQuery does not map repeated query params to arrays
When multiple query parameters with the same key are provided
(`?sellingType=COMPANY&sellingType=KENNITALA`), Nestia's
`@TypedQuery()` decorator always delivers only the first value
(`"COMPANY"`) to the controller instead of the expected array
(`["COMPANY", "KENNITALA"]`). This makes it impossible to build
multi-select filters using unions/enum arrays validated by Typia. The
request is issued by the **generated Nestia SDK** on the frontend, so
the query string matches the SDK's serialization exactly.

## Environment

- `@nestia/core`: `9.1.0`
- `typia`: `10.1.0`
- NestJS 11 (Express platform)
- Node.js 24.11.0

## Reproduction Steps

1. Define query DTO that expects an array:

   ```ts
   export interface BusinessListingFilters {
     sellingType?: SellingType[] & tags.MinItems<1>;
   }
   ```

2. Use it in a controller:

   ```ts
   @TypedRoute.Get()
   async list(@TypedQuery() filters: BusinessListingFilters) {
     console.log(filters.sellingType); // <-- observed value
   }
   ```

3. Call the endpoint (via the generated SDK or manually) with repeated
   params:

   ```
   GET /business-listings?sellingType=COMPANY&sellingType=KENNITALA
   ```

## Expected Behavior

`filters.sellingType` should be an array with both values:

```ts
['COMPANY', 'KENNITALA']
```

## Actual Behavior

Only the first value is returned:

```ts
'COMPANY'
```

The generated controller code (see `dist/...controller.js`) calls
`input.get("sellingType")`, so repeated params are ignored.

- A possible fix would be to call `URLSearchParams.getAll()` for keys
  that are declared as arrays, or to allow a custom mapper to return
  arrays without rewriting the decorator by hand.

Please advise whether array support for repeated query parameters can
be added to `@TypedQuery`, or if there is a recommended pattern for
this use case.

samchon:
## Summary
- add a query fixture matching #1404's enum-array DTO with typia MinItems tag
- assert generated SDK requests with repeated query keys round-trip through @TypedQuery as arrays

## Verification
- pnpm --filter @nestia/sdk run build

## Local gaps
- node tests/test-sdk/start.js --only query --concurrency 1 is blocked on Windows by native transform temp-path normalization and ttsx drive-letter URL handling; CI runs the feature on Ubuntu.

Closes #1404

### DIFF
--- a/tests/test-sdk/features/query/src/api/structures/IBusinessListingFilters.ts
+++ b/tests/test-sdk/features/query/src/api/structures/IBusinessListingFilters.ts
@@ -1,0 +1,9 @@
+import { tags } from "typia";
+
+export interface IBusinessListingFilters {
+  sellingType?: IBusinessListingFilters.SellingType[] & tags.MinItems<1>;
+}
+
+export namespace IBusinessListingFilters {
+  export type SellingType = "COMPANY" | "KENNITALA";
+}
--- a/tests/test-sdk/features/query/src/controllers/QueryController.ts
+++ b/tests/test-sdk/features/query/src/controllers/QueryController.ts
@@ -2,6 +2,7 @@ import { TypedQuery, TypedRoute } from "@nestia/core";
 import { Controller, Query } from "@nestjs/common";
 
 import { IBigQuery } from "@api/lib/structures/IBigQuery";
+import { IBusinessListingFilters } from "@api/lib/structures/IBusinessListingFilters";
 import { INestQuery } from "@api/lib/structures/INestQuery";
 import { IQuery } from "@api/lib/structures/IQuery";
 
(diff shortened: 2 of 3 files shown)


samchon:

1. Reproduction-shape pass: the new fixture uses the reported `sellingType?: Enum[] & tags.MinItems<1>` shape instead of a simplified array, so it pins the branch that regressed in #1404.
2. Runtime-contract pass: the controller echoes the parsed `@TypedQuery()` DTO and the SDK test asserts both repeated query values survive in order, covering the observable client/server behavior.
3. Scope pass: the PR adds only a regression fixture and test; it does not change query-generation logic, so it cannot alter public behavior beyond guarding the existing fix.

GitHub currently attaches only spell-check to this rebased test-only PR and it passed. Local verification also passed `pnpm --filter @nestia/sdk run build`; the full query fixture runtime remains blocked locally on the known Windows native-transform/ttsx path handling issue recorded in the commit.

PR #1487, opened 2026-06-11
fix(migrate): avoid API accessor shadowing

### LINKED ISSUE #1390: `@nestia/migrate` generates duplicated API function name with namespace

samchon:
## Summary
- clone and normalize migrate route accessors at context creation
- escape longer child namespace segments when a shorter route function would share the same exported binding
- add a test-migrate regression that forces a prefix accessor collision and checks functional/e2e output follows the escaped accessor

## Verification
- pnpm --filter @nestia/migrate run build
- pnpm exec ttsc --noEmit -p tests/test-migrate/tsconfig.json

## Local gaps
- pnpm --filter @nestia/test-migrate start is blocked locally on Windows by native transform temp-path normalization during fixture Swagger generation; CI runs the suite on Ubuntu.

Closes #1390

### DIFF
--- a/packages/migrate/src/NestiaMigrateApplication.ts
+++ b/packages/migrate/src/NestiaMigrateApplication.ts
@@ -1,5 +1,6 @@
 import {
   IHttpMigrateApplication,
+  IHttpMigrateRoute,
   OpenApi,
   OpenApiV3,
   OpenApiV3_1,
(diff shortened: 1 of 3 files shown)


samchon:
Self-review after green CI (3 passes):

1. Generator-boundary pass: accessor escaping happens once in `createContext`, before all migrate programmers consume routes, so SDK files and generated e2e calls cannot diverge.
2. Collision-behavior pass: the shorter route keeps the public function name while only the longer child namespace is escaped, preserving the nicer API surface and avoiding duplicate exports.
3. Regression-test pass: the new migrate test forces the exact prefix collision, then checks root export, escaped child file, and generated e2e call path.

CI is green: build/Ubuntu, test/go, test/sdk, test/migrate, test/transform-options, test/e2e, test/benchmark, website/deploy, spell-check, and Socket checks all passed.

PR #1488, opened 2026-06-11
test(core): guard inline Promise return analysis

### LINKED ISSUE #1382: @nestia/core MethodTransformer crashes on inline/utility Promise return types ("get_name" undefined / "invalid promise type")
## Summary

nestia swagger crashes in @nestia/core 8.0.5 when a controller method decorated with @TypedRoute returns a Promise whose inner type is a complex inferred type
(e.g., inline object, utility/conditional types, ReturnType<>, or dynamic import() types). Two failure modes are observed:

1. get_name assumes a declaration exists and crashes:
TypeError: Cannot read properties of undefined (reading '0')
    at get_name (.../@nestia/core/src/transformers/MethodTransformer.ts:92:41)
    at Object.MethodTransformer.transform (.../MethodTransformer.ts:22:58)
2. Promise unwrapping throws:
Error on ImportAnalyzer.analyze(): invalid promise type.
    at .../@nestia/core/src/transformers/MethodTransformer.ts:85:13
These happen during the transformer’s return-type analysis, before Swagger document composition.

## Environment

- @nestia/core: 8.0.5
- @nestia/sdk: 8.0.5
- nestia: 8.0.5
- typescript: 5.7.2 (also reproducible on 5.9.2)
- Framework: NestJS 11.x (application code)

## Steps to Reproduce

Minimal sketches (both reproduce locally):

Inline/anonymous return type:
```
import { TypedRoute } from "@nestia/core";

@Controller("repro")
export class ReproController {
  // When Promise’s inner type is an inline/complex type,
  // symbol resolution can yield no declarations and crash in MethodTransformer.
  @TypedRoute.Get("")
  async repro(): Promise<{ foo: string; bar?: number }> {
    return { foo: "ok" };
  }
}
```
Generic envelope with inline inner type:
```
type Envelope<T> = { message: string; data: T | null };

@Controller("repro2")
export class Repro2Controller {
  @TypedRoute.Get("")
  async repro2(): Promise<Envelope<{ foo: string; bar?: number }>> {
    return { message: "ok", data: { foo: "x" } };
  }
}
```

How to trigger:

1. Use @TypedRoute on the method.
2. Make its return type Promise<...> where ... is an inline/composite/utility/conditional type (in our project, ReturnType<> and dynamic import() also increased
repro rate; they are not strictly required).
3. Run OpenAPI generation:
```
pnpm i
pnpm run openapi:gen
```
The process aborts with one of the stack traces above.

## Expected Behavior

Swagger generation should succeed. If a symbol lacks declarations or a Promise’s type argument cannot be determined, the transformer should safely fall back
instead of throwing.

## Actual Behavior

- get_name(symbol) directly accesses symbol.getDeclarations()[0], which can be undefined/empty for certain symbols (built-ins, synthesized, or complexly inferred
types), causing the TypeError.
- escape_promise(type) assumes checker.getTypeArguments(...) returns exactly one argument and throws otherwise. For conditional/utility/inline types, this can
return zero and throws "invalid promise type".

Note: Because get_name is called before the strict promise check (to compare the symbol name against "Promise"), the first failure mode may occur even when the
return type is not strictly recognized as Promise yet, if the symbol has no declarations.

## Root Cause (analysis)

- Missing declarations: TypeScript symbols for some built-ins or inferred/aliased types legitimately return no declarations. Code that assumes
symbol.getDeclarations()[0] will crash.
- Promise unwrapping assumption: checker.getTypeArguments(...) may legally return 0 for some shapes (or more than one in rare cases). Treating “not exactly 1” as
an error is brittle for complex signatures.

## Proposed Fix (upstream)

Make both helpers defensive:

- get_name(symbol): guard getDeclarations() and fall back to escapedName when empty.
- escape_promise(type): do not throw when no type arguments are present; return the original type (or the first argument if present). Prefer safe unwrapping
rather than hard failure.

Working patch (TypeScript):
```
--- a/src/transformers/MethodTransformer.ts
+++ b/src/transformers/MethodTransformer.ts
@@
-const get_name = (symbol: ts.Symbol): string =>
-  explore_name(symbol.getDeclarations()![0].parent)(
-    symbol.escapedName.toString(),
-  );
+const get_name = (symbol: ts.Symbol): string => {
+  const decls: readonly ts.Declaration[] | undefined =
+    (symbol as any).getDeclarations?.();
+  const name = (symbol as any).escapedName?.toString?.() ?? "";
+  if (!decls?.length) return name;
+  const parent: ts.Node = (decls[0] as any)?.parent ?? decls[0];
+  return explore_name(parent)(name);
+};

@@
 const escape_promise =
   (checker: ts.TypeChecker) =>
   (type: ts.Type): ts.Type => {
-    const generic: readonly ts.Type[] = checker.getTypeArguments(
-      type as ts.TypeReference,
-    );
-    if (generic.length !== 1)
-      throw new Error(
-        "Error on ImportAnalyzer.analyze(): invalid promise type.",
-      );
-    return generic[0];
+    try {
+      const generic: readonly ts.Type[] =
+        checker.getTypeArguments(type as ts.TypeReference);
+      if (!generic?.length) return type;
+      return generic[0] ?? type;
+    } catch {
+      return type;
+    }
   };
```

For local verification we applied the same defensive changes to the distributed lib/transformers/MethodTransformer.js as well. With these guards, Swagger
generation completes successfully.

## Workarounds (project-side)

- Prefer named DTO classes/interfaces as return types at route boundaries instead of inline/anonymous composites.
- Avoid ReturnType<>, dynamic import(), or stacked utility/conditional types directly in the immediate Promise<...> position (wrap them in a named DTO instead).

## Additional Context

- Public repo: https://github.com/Neuron-Grid/Mycelia-API
- Repro commands:
  pnpm i
  pnpm run openapi:gen
- With the above defensive patch applied, generation completes (observed metrics in our app: controllers: 15, paths: 79, routes: 79).

If the direction looks acceptable, I’m happy to put together a PR (including tests around inline/utility return types and promise unwrapping behavior). Also, a
separate (possibly follow-up) topic is that generic envelopes (e.g., Envelope<T>) may produce {} schemas in OpenAPI output even when generation succeeds; that
appears orthogonal to the transformer crash and can be filed independently.

samchon:
## Summary
- add a Go regression test for inline object Promise returns and generic Promise envelopes
- pin the native return-type helper behavior that replaces the old TypeScript MethodTransformer crash path from #1382

## Verification
- go test ./... -run TestMethodReturnTypeInlinePromise -count=1 from packages/core/test using bundled Go

## Local gaps
- full packages/core/test suite is blocked locally on Windows by native transform temp-path normalization in transform fixture tests; CI runs the suite on Ubuntu.

Closes #1382

### DIFF
--- a/packages/core/test/method_return_type_inline_promise_test.go
+++ b/packages/core/test/method_return_type_inline_promise_test.go
@@ -1,0 +1,98 @@
+package test
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	shimast "github.com/microsoft/typescript-go/shim/ast"
+	"github.com/samchon/nestia/packages/core/native/transform"
+	"github.com/samchon/ttsc/packages/ttsc/driver"
+)
+
+// TestMethodReturnTypeInlinePromise verifies inline and generic Promise return
+// types do not crash return-type analysis.
+//
+// Locks the native replacement for the old TypeScript MethodTransformer branch
+// reported in #1382. Inline object literals and generic envelopes can expose
+// synthesized symbols, so the helper must unwrap Promise<T> only when the type
+// argument is available and otherwise fall back without dereferencing missing
+// declarations.
+//
+//  1. Load a temporary program with inline and generic Promise return types.
+//  2. Visit both method declarations.
+//  3. Assert NestiaCoreMethodReturnType returns a type for each method without
+//     panicking.
+func TestMethodReturnTypeInlinePromise(t *testing.T) {
+	temp := t.TempDir()
+	writeFile(t, filepath.Join(temp, "tsconfig.json"), `{
+  "compilerOptions": {
+    "target": "ES2020",
+    "module": "CommonJS",
+    "strict": true
+  },
+  "include": ["controller.ts"]
+}`)
+	writeFile(t, filepath.Join(temp, "controller.ts"), `
+type Envelope<T> = { message: string; data: T | null };
+
+class ReproController {
+  public async inline(): Promise<{ foo: string; bar?: number }> {
+    return { foo: "ok" };
+  }
+
+  public async envelope(): Promise<Envelope<{ foo: string; bar?: number }>> {
+    return { message: "ok", data: { foo: "ok" } };
+  }
+}
+`)
+
+	prog, diags, err := driver.LoadProgram(temp, "tsconfig.json", driver.LoadProgramOptions{ForceNoEmit: true})
+	if err != nil {
+		t.Fatalf("LoadProgram failed: %v", err)
+	}
+	if len(diags) > 0 {
+		t.Fatalf("LoadProgram produced %d config diagnostics", len(diags))
+	}
+	defer prog.Close()
+
+	source := prog.SourceFile(filepath.ToSlash(filepath.Join(temp, "controller.ts")))
+	if source == nil {
+		t.Fatal("controller.ts not in program")
+	}
+
+	seen := map[string]bool{}
+	var walk func(node *shimast.Node)
+	walk = func(node *shimast.Node) {
+		if node == nil {
+			return
+		}
+		if node.Kind == shimast.KindMethodDeclaration {
+			name := transform.NodeName(node)
+			if name == "inline" || name == "envelope" {
+				if typ := transform.NestiaCoreMethodReturnType(prog, node); typ == nil {
+					t.Fatalf("%s return type was not resolved", name)
+				}
+				seen[name] = true
+			}
+		}
+		node.ForEachChild(func(child *shimast.Node) bool {
+			walk(child)
+			return false
+		})
+	}
+	walk(source.AsNode())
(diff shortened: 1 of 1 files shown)


samchon:

1. Reproduction-shape pass: the Go test covers both inline object `Promise<{ ... }>` and generic envelope `Promise<Envelope<{ ... }>>` return types, matching the crash family reported in #1382.
2. Native-helper pass: the test exercises `NestiaCoreMethodReturnType` through a loaded TypeScript program and asserts the helper returns a type without panicking for both methods.
3. Scope pass: this is a regression test only; current native code is already defensive, so no runtime or public API behavior changes are introduced.

Checks are green: website/deploy, spell-check, and Socket checks passed. Local verification also passed `go test ./... -run TestMethodReturnTypeInlinePromise -count=1` from `packages/core/test` on this branch.

PR #1490, opened 2026-06-11
test(sdk): call query leaf accessor

samchon:
## Summary
- call the generated `typed_enum_array.typedEnumArray` leaf function in the repeated query regression test
- fix the full `tests/test-sdk` query feature typecheck failure exposed by downstream CI

## Verification
- `git diff --check`
- `pnpm exec prettier --check tests/test-sdk/features/query/src/test/features/api/test_api_query_typed_repeated_enum_array.ts`

## Notes
- Local `tests/test-sdk` query runtime remains blocked on Windows by `spawn EINVAL` before feature execution. The downstream CI log shows the generated `typed_enum_array` namespace and the lack of call signature; this PR calls the generated leaf function under that namespace.

Related: #1404

### DIFF
--- a/tests/test-sdk/features/query/src/test/features/api/test_api_query_typed_repeated_enum_array.ts
+++ b/tests/test-sdk/features/query/src/test/features/api/test_api_query_typed_repeated_enum_array.ts
@@ -23,7 +23,10 @@ export const test_api_query_typed_repeated_enum_array = async (
     sellingType: ["COMPANY", "KENNITALA"],
   };
   const result: IBusinessListingFilters =
-    await api.functional.query.typed_enum_array(connection, input);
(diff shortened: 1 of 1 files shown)


samchon:

1. Accessor-shape pass: the route accessor is `query.typed_enum_array.typedEnumArray`, matching the path-segment namespace plus controller-method leaf that the SDK generator emits.
2. Failure-log pass: this directly addresses the CI error that `typed_enum_array` is a namespace with no call signatures by calling the leaf function inside it.
3. Scope pass: only the regression test call site changes; production code and fixture route behavior are untouched.

Checks are green: Socket project report, Socket PR alerts, and spell-check passed. Local verification also passed `git diff --check`, Prettier check for the touched file, and `pnpm --filter @nestia/sdk run build`.

Issue #1186, opened 2025-01-12
NestiaSwaggerComposer.compose(): Unable to find exception type.


chigix:
### Summary
Write a short summary of the bug in here.

  - **SDK Version**: 4.5.2
  - **Expected behavior**: `@TypedException<BadRequestException>({ status: 400, description: 'Invalid parameter provided.' })` could find the exception type and generate Error case in Swagger document.
  - **Actual behavior**: This `@TypedException` triggers `NestiaSwaggerComposer.compose()` erroring "Unable to find exception type".

Write detailed description in here.

### Code occuring the bug

Referring to the case writing style from https://github.com/samchon/nestia/pull/548 , below is a shorter case that returns datetime where 400 is designated for invalid parameter request.

```typescript
class HttpException extends Error {}

@Controller('get-datetime')
@ApiTags('Datetime')
export class DatetimeController {
  /**
   * /api/get-datetime
   */
  constructor(
    private readonly datetimeService: DatetimeService,
  ) { }

  /**
   * Retrieves the current date and time.
   *
   * @summary Retrieve Current Datetime
   *
   * @param timezone The timezone identifier (e.g., "UTC", "America/New_York").
   * @returns The current date and time in the specified timezone, in ISO 8601 format.
   */
  @Get()
  @Header('Content-Type', 'text/plain')
  @TypedException<HttpException>({
    status: 400,
    description: 'Invalid parameter provided.',
  })
  public getDateTime(
    @Query('timezone') timezone?: string,
  ): string {
    return this.datetimeService.getCurrentDatetime(timezone);
  }
}
```

I would always get the following error log:
```
.../node_modules/@nestia/sdk/src/NestiaSwaggerComposer.ts:134
  throw new Error(`Error on NestiaSwaggerComposer.compose():\n${messages}`);
        ^
Error: Error on NestiaSwaggerComposer.compose():
dist/apps/.../main.js - DatetimeController.getDateTime() from exception (status: 400):
  - Unable to find exception type.
    at report (.../node_modules/@nestia/sdk/src/NestiaSwaggerComposer.ts:134:9)
    at analyze (.../node_modules/@nestia/sdk/src/NestiaSwaggerComposer.ts:55:13)
    at .../node_modules/@nestia/sdk/src/NestiaSwaggerComposer.ts:30:15
    at Generator.next (<anonymous>)
    at fulfilled (.../node_modules/@nestia/sdk/lib/NestiaSwaggerComposer.js:5:58)
```

[...]

PR #1491, opened 2026-06-11
test(sdk): cover NestJS exception classes

samchon:
## Summary
- add an exception fixture route using `@TypedException<BadRequestException>` from `@nestjs/common`
- assert Swagger emits the 400 response description and `BadRequestException` schema reference
- extend the SDK native metadata in-process test to require the imported NestJS exception class metadata

## Verification
- `go test ./... -run TestSDKExceptionMetadataInProcess -count=1` from `packages/sdk/test`
- `git diff --check`

## Notes
- Local full `tests/test-sdk` runtime remains blocked by Windows test runner/native path issues, so CI is the authoritative full-suite verification.

Related: #1186

### DIFF
--- a/packages/sdk/test/sdk_exception_metadata_inprocess_test.go
+++ b/packages/sdk/test/sdk_exception_metadata_inprocess_test.go
@@ -43,7 +43,8 @@ func TestSDKExceptionMetadataInProcess(t *testing.T) {
 		`"name":"TypeGuardErrorany"`,
 		`"value":"method"`,
 		`"value":"expected"`,
-		// The four named exception structures collected across decorators.
+		// The named exception structures collected across decorators.
+		`"name":"BadRequestException"`,
 		`"name":"INotFound"`,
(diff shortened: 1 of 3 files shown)


samchon:

1. Issue-shape pass: the fixture now uses `@TypedException<BadRequestException>` imported from `@nestjs/common`, matching the failure shape reported in #1186 instead of only local DTO/interface exceptions.
2. Metadata pass: the SDK native in-process test now requires `BadRequestException` metadata, so the transform must emit the imported NestJS exception class before Swagger composition can consume it.
3. Swagger pass: the generated Swagger fixture asserts both the 400 description and `#/components/schemas/BadRequestException` response schema for the NestJS exception route.

Checks are green: website/deploy, spell-check, and Socket checks passed. Local verification also passed `pnpm --filter @nestia/sdk run build` and `go test ./... -run TestSDKExceptionMetadataInProcess -count=1` from `packages/sdk/test`.

PR #1492, opened 2026-06-11
feat(sdk): support Observable route returns

### LINKED ISSUE #747: Support Observable return types for Nestia SDK
## Feature Request
When using the Nestia Client SDK, it can properly infer the controller method return value types when returning synchronously or returning a Promise. However, if we try to return an Observable, then the SDK is not able to generate types properly.

Since NestJS has first-class support for Observables, many times we work with Observables exclusively and would just like to return the Observable from the controller method instead of converting it to a Promise and then returning the Promise. However, returning an Observable type breaks Nestia's ability to properly type the response type.

Workaround's we need to use today:
1. Use async / await and only work with Promises
2. Use the `firstValueFrom` rxjs operator to convert Observable to Promise before returning the response from the controller.

This works fine:
```typescript
   @TypedRoute.Get('user/:id')
   async getUser(@Param('id') id: string): Promise<User> {
     return firstValueFrom(this.userService.findById(id));
   }
   ````

This does not:
```typescript
   @TypedRoute.Get('user/:id')
   getUser(@Param('id') id: string): Observable<User> {
     return this.userService.findById(id);
   }
   ````

samchon:
## Summary
- unwrap explicit `Observable<T>` route return annotations in the native core return-type helper
- mirror Promise/Observable unwrapping in SDK type-node reflection helpers
- add Go and test-sdk route regression coverage for generated SDK payload typing

## Verification
- `go test ./... -run TestMethodReturnTypeObservable -count=1` from `packages/core/test`
- `pnpm exec prettier --check tests/test-sdk/features/route/src/controllers/TypedRouteController.ts tests/test-sdk/features/route/src/test/features/api/test_api_route_observable.ts`
- `git diff --check`

## Notes
- Local `tests/test-sdk` route runtime is blocked on Windows by `spawn EINVAL` before feature execution. CI should run the Linux fixture suite.

Fixes #747

### DIFF
--- a/packages/core/native/transform/core_transform.go
+++ b/packages/core/native/transform/core_transform.go
@@ -1513,6 +1513,9 @@ func cleanupNestiaCorePrintedExpression(text string) string {
 var nestiaCoreSingleParameterArrowPattern = regexp.MustCompile(`(^|[\s(=,:?])([A-Za-z_$][A-Za-z0-9_$]*) =>`)
 
 func NestiaCoreMethodReturnType(prog *driver.Program, node *shimast.Node) *shimchecker.Type {
+	if typ := nestiaCoreExplicitAsyncReturnType(prog, node); typ != nil {
+		return typ
+	}
 	signature := prog.Checker.GetSignatureFromDeclaration(node)
 	if signature == nil {
 		return nil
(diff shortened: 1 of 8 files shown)


samchon:
Self-review after green CI/checks (3 passes):

1. Native transform pass: re-read the core return-type helper and SDK type-node helpers. The change keeps async wrapper handling centralized and extends the existing Promise<T> policy to Observable<T> without reintroducing a TypeScript-side transformer.
2. Fixture/dependency pass: re-read the route fixture, generated SDK test, and workspace dependency changes. The route uses real RxJS Observable/of because Nest's runtime Observable handling depends on RxJS semantics; `rxjs` is scoped to `tests/test-sdk` through the `nestjs` catalog and the lockfile only adds the importer/catalog entries for an already resolved version.
3. Verification pass: CI is green for Ubuntu, deploy, go, sdk, e2e, migrate, benchmark, transform-options, spell-check, and Socket Project Report. Socket PR Alerts is skipped by provider behavior, not a failing check. Local verification also covered the targeted Go test, SDK package build, route fixture Prettier check, and `git diff --check`; the Windows route runner still cannot start locally due `spawn EINVAL` before feature execution.

PR #1496, opened 2026-06-12
fix(sdk): expose readonly array OpenAPI extension

### LINKED ISSUE #1403: ReadOnly flag missing in OpenApi Schema
## Bug Report
After updating to the new version 9.1.0 the readonly flag is not being set in my test project.
It only detects the readOnly flag if set to the parameter itself, like:

```ts
  readonly options: SelectMenuComponentOptionData[];
```

but not if it is set to the type, like: 
```ts
 options: readonly SelectMenuComponentOptionData[];
```
OpenApi doesn't have a real way to handle this, maybe have to work with x-readonly-array 

This is the openapi result:
```json
          "options": {
            "type": "array",
            "items": {
              "$ref": "#/components/schemas/SelectMenuComponentOptionData"
            }
          },
```

with this route

```ts
          @core.TypedRoute.Post(":id")
          public index(
            @core.TypedParam("id") id: string,
            @core.TypedBody() data: StringSelectMenuComponentData,
          ): Promise<StringSelectMenuComponentData> {
            return ServersProvider.setData({
              section: id,
              input: data,
            });
          }
```
and this type
```ts
export interface StringSelectMenuComponentData extends BaseSelectMenuComponentData {
  type: ComponentType.StringSelect;
  options: readonly SelectMenuComponentOptionData[];
}
```

```json
    "@nestia/benchmark": "^9.1.0",
    "@nestia/e2e": "^9.1.0",
    "@nestia/sdk": "^9.1.0",
    "nestia": "^9.1.0",
```

samchon:
## Summary

- add native SDK schema-bake handling that annotates TypeScript `readonly T[]` / `ReadonlyArray<T>` schema positions with `x-readonly-array: true`
- keep OpenAPI `readOnly` reserved for readonly object properties so response-only semantics are not conflated with array immutability
- guard the native readonly-array schema walker against recursive DTO graphs, handle both interface and type-alias DTO components, preserve the vendor extension through Swagger generation, cover it with a generated-Swagger `swagger-customizer` regression test, and document the extension

Fixes #1403.

## Verification

- `pnpm exec prettier --check packages/sdk/src/generates/SwaggerGenerator.ts packages/sdk/src/generates/internal/SwaggerOperationParameterComposer.ts packages/sdk/src/generates/internal/SwaggerReadonlyArrayEmender.ts tests/test-sdk/features/swagger-customizer/src/controllers/SwaggerController.ts tests/test-sdk/features/swagger-customizer/src/test/features/test_swagger_readonly_array.ts`
- `pnpm exec ttsc --noEmit --project tests/test-sdk/features/swagger-customizer/tsconfig.json`
- `git diff --check`

## Local limitation

- `gofmt` and direct `go test` are unavailable locally because the Go toolchain is not on PATH; `ttsc --noEmit` did compile the native source-plugin contributor.
- `node tests/test-sdk/start.js --only clone-and-keyword --concurrency 1` and `node tests/test-sdk/start.js --only swagger-customizer --concurrency 1` currently fail on local Windows before generated tests run with native TypeScript-Go temp-file path normalization: `fileName should be normalized and absolute: "C:\\Users\\samch\\AppData\\Local\\Temp\\...ts"`.

### DIFF
--- a/packages/sdk/native/sdk/sdk_transform.go
+++ b/packages/sdk/native/sdk/sdk_transform.go
@@ -17,6 +17,7 @@ import (
 	"github.com/samchon/ttsc/packages/ttsc/driver"
 	nativecontext "github.com/samchon/typia/packages/typia/native/core/context"
 	nativefactories "github.com/samchon/typia/packages/typia/native/core/factories"
+	nativeiterate "github.com/samchon/typia/packages/typia/native/core/programmers/iterate"
 	nativejson "github.com/samchon/typia/packages/typia/native/core/programmers/json"
 	schemametadata "github.com/samchon/typia/packages/typia/native/core/schemas/metadata"
 )
@@ -704,7 +705,7 @@ func nestiaSDKSchemaPipe(context *nestiaSDKContext, typ *shimchecker.Type, typeN
 	// and omit the field — the sdk generator falls back to its own derived
 	// schema path. This must never mask a real bug, so re-raise anything we
 	// don't recognize as a transformer error from the typia runtime.
-	if baked := nestiaSDKTryBakeJsonSchema(result.Data); baked != nil {
+	if baked := nestiaSDKTryBakeJsonSchema(prog, typeNode, result.Data); baked != nil {
 		metadataLiteral["jsonSchema"] = baked
 	}
 	value := map[string]any{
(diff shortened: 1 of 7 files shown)


samchon:
Post-CI self-review completed after the final green run on `d8b1957670cace2b31cb38dfa9b45581e3526bf0`.

Pass 1 - native schema bake: checked readonly array marking for `readonly T[]`, `ReadonlyArray<T>`, nested array/tuple items, interface/type-alias component `$ref` resolution, and recursive DTO traversal. The current traversal stack guard avoids the prior clone-fixture stack overflow without globally suppressing reused aliases.

Pass 2 - Swagger generation paths: checked the top-level schema collection path and decomposed query/header parameter path. The emender preserves `x-readonly-array` while keeping OpenAPI `readOnly` reserved for readonly object properties.

Pass 3 - tests/docs/release surface: checked generated-Swagger coverage for interface and type-alias DTOs, including mutable arrays, readonly arrays, readonly properties, and combined readonly cases. Docs explain the vendor extension and the PR title/body match the release intent.

All CI checks are passing; no remaining blocking findings.

PR #1497, opened 2026-06-12
fix(core): preserve TypedRoute JSON content type

### LINKED ISSUE #766: Content type of response changes randomly to text/plain (instead of application/json)
## Question
I use @TypedRoute.Get() in a controller to retrieve data. Usually I get as the expected content-type application/json but in random responses I get text/plain as content-type. If I use the default Nest.js decorator @Get() instead of @TypedRoute.Get(), the response always consistently has the expected content-type application/json. Is there an explanation for this behavior?

samchon:
Fixes #766.

## Summary
- Register Nest route-level `Content-Type: application/json` metadata on every `@TypedRoute` decorator.
- Add a route feature regression where an outer interceptor returns pre-serialized cached JSON without calling the next handler.
- Assert the raw HTTP response still reports `application/json` instead of falling back to `text/plain`.

## Root Cause
`TypedRouteInterceptor` already writes the JSON content type, but cache-like interceptors can short-circuit the request before that interceptor path runs. When the response body is already serialized text and no route-level header metadata exists, the HTTP adapter can infer `text/plain`.

## Regression Shape
The fixture keeps the cache interceptor before `TypedRouteInterceptor` in Nest metadata order, matching the short-circuit path. This lets the generated SDK test parse the cached JSON body while the dedicated raw HTTP test verifies the response header.

## Verification
- `pnpm exec prettier --check packages/core/src/decorators/TypedRoute.ts tests/test-sdk/features/route/src/controllers/TypedRouteController.ts tests/test-sdk/features/route/src/test/features/api/test_api_route_cached_content_type.ts`
- `git diff --check`
- `pnpm --filter @nestia/core build`
- `pnpm --filter @nestia/sdk build` after `@nestia/core` build completed
- Direct metadata check against `packages/core/lib` confirmed the cache interceptor precedes `TypedRouteInterceptor` and `HEADERS_METADATA` contains `Content-Type: application/json`.

## Local Test Limit
- `node tests/test-sdk/start.js --only route --concurrency 1` is blocked on this Windows checkout by the existing TypeScript-Go native temp-path normalization failure before the route regression test runs: `fileName should be normalized and absolute: "C:\Users\samch\AppData\Local\Temp\...ts"`. The Linux CI `test-sdk` job should exercise the new regression end to end.

### DIFF
--- a/packages/core/src/decorators/TypedRoute.ts
+++ b/packages/core/src/decorators/TypedRoute.ts
@@ -3,6 +3,7 @@ import {
   Delete,
   ExecutionContext,
   Get,
+  Header,
   NestInterceptor,
   Patch,
   Post,
@@ -141,6 +142,7 @@ export namespace TypedRoute {
       )(...args);
       return applyDecorators(
         ROUTERS[method](path),
+        Header("Content-Type", "application/json"),
         UseInterceptors(new TypedRouteInterceptor(stringify)),
       );
     }
--- a/tests/test-sdk/features/route/src/controllers/TypedRouteController.ts
+++ b/tests/test-sdk/features/route/src/controllers/TypedRouteController.ts
@@ -1,5 +1,11 @@
 import core from "@nestia/core";
-import { Controller } from "@nestjs/common";
+import {
+  CallHandler,
+  Controller,
+  ExecutionContext,
+  NestInterceptor,
+  UseInterceptors,
+} from "@nestjs/common";
 import { Observable, of } from "rxjs";
 import typia from "typia";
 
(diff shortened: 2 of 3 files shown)


samchon:
Self-review after green CI:

1. Pass 1 found that the raw header regression test could still pass on a JSON 500 response. Fixed by requiring `response.ok` and an object payload with the cached article id.
2. Pass 2 rechecked runtime and generator impact: `@TypedRoute` now mirrors its existing interceptor JSON header as Nest route metadata, and the SDK response analyzer treats that as the same JSON success content type.
3. Pass 3 rechecked final diff, PR metadata, branch target, and CI. No remaining blockers found.

Final CI on `4f98d328ab9f184f5c6bf1ea696bd8db686014e6`: Ubuntu, deploy, go, benchmark, e2e, migrate, sdk, spell-check, transform-options, and Socket checks all passed.

PR #1502, opened 2026-06-12
feat(sdk): add swagger watch mode

### LINKED ISSUE #703: Add swagger generation with --watch mode / live reload
Please add a way to support live reloading with nestia

For instance in our project we use the swagger.json file 

```ts
import swaggerDocument from '../swagger.json';

SwaggerModule.setup('/api', app, swaggerDocument as OpenAPIObject, {
    swaggerOptions: {
      defaultModelRendering: 'model',
      tryItOutEnabled: true,
      syntaxHighlight: {
        activate: true,
      },
    },
});
 ```
 
Then we run the ``nestia swagger`` command to regenerate the schema, but it would be nice to have nestia to watch for changes and re generate the schema when a controller change.

This would be a really nice 'ease of use' feature.

I guess we could try to make a webpack plugin that run the command when a controller change but i think it would be better if it was handled by nestia itself.

Thanks again for this great tool <3

samchon:
﻿## Summary

Closes #703.

Adds `npx nestia swagger --watch` / `npx @nestia/sdk swagger --watch` for local live reload workflows. The command runs the normal swagger generator immediately, then watches controller inputs, `nestia.config.ts`, and `tsconfig.json`-style project inputs and reruns the same generator after a debounced file change.

## Implementation

- adds `NestiaSdkWatcher` using Node `fs.watch` and the existing `glob` package, with no new dependencies
- keeps one-shot generation and watch generation on the same `NestiaSdkApplication.swagger()` path
- ignores generated SDK/swagger output paths, `node_modules`, `.git`, and `.nestia` so output writes do not loop the watcher
- adds a `swagger-watch` synthetic test-sdk feature that starts the real CLI watch process, writes an isolated temporary fixture, edits its controller, and waits for `swagger.json` to gain the new route
- documents `nestia swagger --watch` in the tutorial and swagger reference pages

## Validation

- `pnpm --filter nestia build`
- `TEST_SDK_SKIP_BUILD=1 node tests/test-sdk/start.js --only swagger-watch --concurrency 1`
- `pnpm exec prettier --check packages/sdk/src/executable/internal/NestiaSdkWatcher.ts packages/sdk/src/executable/internal/NestiaSdkCommand.ts packages/sdk/src/executable/sdk.ts packages/cli/src/index.ts tests/test-sdk/start.js`
- `git diff --cached --check`

## Local limitations

- `node tests/test-sdk/start.js --only swagger --concurrency 1` still fails locally on existing Windows tsgo temp-path normalization issues unrelated to this change. The new `swagger-watch` synthetic case passes when executed alone.
- `npm --prefix website run build` fails before docs compile because `website/build/editor.js` cannot resolve `jszip` in this checkout.
- `pnpm exec prettier --check --parser mdx ...` warns on the touched MDX files because the existing docs formatting does not match Prettier output; the MDX change is intentionally minimal.

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -9,7 +9,7 @@ npx nestia [command] [options?]
   4. npx nestia dependencies
   5. npx nestia init
   6. npx nestia sdk
-  7. npx nestia swagger
+  7. npx nestia swagger [--watch]
   8. npx nestia e2e
   9. npx nestia all
 `;
@@ -32,9 +32,9 @@ async function main(): Promise<void> {
       await import("./NestiaTemplate.js")
     ).NestiaTemplate.clone((msg) => halt(msg ?? USAGE))(argv);
   } else if (type === "setup") {
-    await (await import("./NestiaSetupWizard.js")).NestiaSetupWizard.setup(
-      argv,
-    );
+    await (
+      await import("./NestiaSetupWizard.js")
+    ).NestiaSetupWizard.setup(argv);
   } else if (
     type === "dependencies" ||
     type === "init" ||
--- a/packages/sdk/src/executable/internal/NestiaSdkCommand.ts
+++ b/packages/sdk/src/executable/internal/NestiaSdkCommand.ts
@@ -1,6 +1,7 @@
 import { INestiaConfig } from "../../INestiaConfig";
 import { NestiaSdkApplication } from "../../NestiaSdkApplication";
 import { NestiaConfigLoader } from "./NestiaConfigLoader";
+import { NestiaSdkWatcher } from "./NestiaSdkWatcher";
 
 export namespace NestiaSdkCommand {
   export const sdk = () =>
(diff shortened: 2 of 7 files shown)


samchon:
﻿Self-review after green CI:

1. Watcher behavior review: confirmed watch mode calls the same NestiaSdkApplication.swagger path as one-shot generation, debounces concurrent changes, ignores generated outputs to avoid self-trigger loops, and reconfigures watched directories after regeneration so newly added controller directories are covered.
2. Test isolation review: replaced fixture-copy based testing with an isolated temporary fixture to avoid parallel test races, and excluded stale `.tmp-*` directories from feature discovery. The watch process is killed and the temp directory is removed in `finally`.
3. CLI/docs review: public surface is limited to `swagger --watch`; `sdk`, `e2e`, and `all` command behavior is unchanged. Usage text and Swagger docs now mention the new mode.

All PR CI checks passed on head `729b5c3f895352e8594c289aca590e7c409f1020`. No remaining changes found from the review passes.

PR #1506, opened 2026-06-18
Drop vendored @nestia/factory for published @ttsc/factory

samchon:
## Why

`@nestia/factory` was an in-repo copy of the TypeScript AST factory/printer that [`@ttsc/factory`](https://www.npmjs.com/package/@ttsc/factory) now ships as a standalone, dependency-free package. Maintaining a fork of upstream code is pure duplication, so this **deletes `packages/factory`** and points `@nestia/sdk` and `@nestia/migrate` at `@ttsc/factory` directly.

## What changed

`@ttsc/factory` is **not** a drop-in for the vendored copy — the representation differs:

- Node kinds are **string** discriminants (`"Identifier"`) instead of numeric `SyntaxKind`.
- `SyntaxKind`/`NodeFlags` are **string enums** covering only tokens/keywords/trivia (not node kinds).
- Synthesized comments live in a `WeakMap` via `addSyntheticLeadingComment`, not on `node.emitNode`.
- Node types are a **precise discriminated union** instead of one permissive `Node`.

So the consumers were ported, not just re-pointed:

- **Canonical imports** everywhere: `import { factory, SyntaxKind, type Node } from "@ttsc/factory"` + `factory.createX(...)`. No `import * as ts`, no `TypeScriptFactory` alias.
- The nestia-only helpers (`IdentifierFactory`, `LiteralFactory`, `ExpressionFactory`, `StatementFactory`, `TypeFactory`, `FormatCheatSheet`) and the printer usage moved into each package's own `src/factories/`, re-implemented against `@ttsc/factory`'s string-kind model. Duplicated per package because `sdk` and `migrate` don't depend on each other.
- AST value annotations were **narrowed** from the catch-all `Node` to the precise `@ttsc/factory` types (`Expression`/`Statement`/`TypeNode`/…) the values actually are.
- `migrate`'s `internal/ts` shim keeps its type aliases (mapped to the precise types) + the `isUnionTypeNode` guard; `factory`/`SyntaxKind`/`NodeFlags` are imported directly.
- Removed `@nestia/factory` from `packages/sdk` & `packages/migrate` deps (added `@ttsc/factory` via the `typescript` catalog), bumped the `@ttsc` catalog to `^0.15.9`, and cleaned up references in `tests/test-sdk/start.js`, `website/package.json`, `AGENTS.md`, and `packages/core/MIGRATION.md`.

## Verification

- `pnpm build` over the full workspace (including the `@nestia/core` Go-backed native transform) is green on `@ttsc/factory@0.15.9`.
- Generated SDK output is unchanged: `@ttsc/factory` ships real runtime `SyntaxKind`/`NodeFlags` objects, and `FilePrinter` still runs prettier over the printer output.
- Not run locally: the `tests/test-*` e2e suites — relying on CI.


samchon:
⚠️ **Blocked on upstream printer fix:** [samchon/ttsc#241](https://github.com/samchon/ttsc/issues/241).

The migration itself is complete and every CI job passes **except `sdk`**, which fails only because clone-mode e2e generation hits a `@ttsc/factory` `TsPrinter` bug: a nested `IntersectionTypeNode` is printed with a dangling `& &` (invalid, `TS1110`), and unions inside intersections aren't parenthesized. The legacy `@nestia/factory` printer had a parenthesizer; `@ttsc/factory`'s doesn't. Per the no-workaround policy, the fix belongs in the printer, not the nestia generator. Once `@ttsc/factory` ships the parenthesizer fix, bump the catalog pin here and the `sdk` job goes green.

PR #1516, opened 2026-06-25
fix(e2e): error handling in promise resolution

sunrabbit123:
# Why

In general, people expect 'Error' to come from the 'catch' statement.
but, current code does throw the `string`

# What Changed

just add `new Error` statement
just one line modefied


samchon (approved):
Thanks for contribution

PR #1528, opened 2026-07-09
fix: restore CI — pin template clones & fix typia ordered-object JSON leak

samchon:
## Why

Every CI job on master is red, from two independent defects that landed today. This PR carries both repairs because neither can show a green check run alone (the install breakage masks the swagger breakage).

### 1. Install breakage — unpinned template clones

`packages/migrate`'s `prepare` script clones `samchon/nestia-start` at its default branch, and [nestia-start#632](https://github.com/samchon/nestia-start/pull/632) restructured that repo into a pnpm monorepo — the single-package paths `bundle.js` deletes no longer exist, so `fs.rm` throws ENOENT and aborts `pnpm install` in every job.

`bundle.js` now checks out an explicit reviewed revision after cloning:
- **nestia-start** → `1057bccd` (last single-package commit; adopting the monorepo template is a separate work item)
- **nestia-sdk-template** → `8cfb771e` (current master, purely for hermeticity)

`fs.rm(..., { force: true })` was rejected — it would silently bundle an incompatible template instead of failing loudly.

### 2. sdk suite breakage — typia ordered-object JSON leak

The typia native bump in #1526 pulled in `LiteralFactory_OrderedObject` (typia `bdf81fde8`, LLM property-order fix). nestia's SDK transform serializes route metadata with `encoding/json`, and the struct had no `MarshalJSON`, so its raw `{Keys, Values}` fields leaked into generated swagger documents as `properties.Keys = [...]` — failing `openapi_v2`, `openapi_v3`, `swagger-customizer`, and `swagger-example` on every run since (#1526's own test run was already red).

Root fix in typia: [samchon/typia@fd1c66869](https://github.com/samchon/typia/commit/fd1c6686950f30357be2b3c484675add2dd7f3c6) adds a `MarshalJSON` override (+ Go regression test). This PR bumps all four Go modules to it.

## Verification

`go build ./...` passes locally in both native modules; full suite verification is this PR's CI.

https://claude.ai/code/session_011JjfmhonWb4ZKAniYipc9e

### DIFF
--- a/packages/core/native/go.mod
+++ b/packages/core/native/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/microsoft/typescript-go/shim/printer v0.0.0
 	github.com/microsoft/typescript-go/shim/scanner v0.0.0
 	github.com/samchon/ttsc/packages/ttsc v0.0.0
-	github.com/samchon/typia/packages/typia/native v0.0.0-20260709031533-27f105ae8e56
+	github.com/samchon/typia/packages/typia/native v0.0.0-20260709134626-fd1c6686950f
 )
 
 require (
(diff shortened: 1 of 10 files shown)


samchon:
Third repair pushed: the `swagger-customizer` failure was a separate regression from the same #1526 dependency bump — the new compiler shim resolves generic type arguments, so array identity names now reach the readonly-array emender replacer-mangled (`ReadonlyArray<string>` → `ReadonlyArraystring`) and none of its literal patterns matched. The emender now matches the prefixes that survive every observed name form. Verified locally by regenerating the fixture: all ten `x-readonly-array`/`readOnly` expectations hold.

PR #1534, opened 2026-07-09
fix(sdk): stop forcing npm into user projects from distribute and dependencies

samchon:
## Summary

`@nestia/sdk` has two surfaces that execute package-manager commands inside the **user's** project, and both hard-coded npm while the ecosystem templates are now pnpm monorepos with `catalog:` deps:

1. **`SdkDistributionComposer`** (runs when `distribute` is configured) executed `npm install ...` in the distribute directory — planting a rogue `package-lock.json` inside pnpm/yarn projects — and unconditionally `copyFile`'d the bundle assets over whatever the user had there.
2. **`nestia dependencies`** defaulted to npm and installed `@nestia/e2e`, `@nestia/fetcher`, and `typia` **unpinned**, so registry latest could pull an incompatible major next to an older SDK.
3. **`TtscExecutor`** error guidance suggested `npm install --save-dev ttsc` regardless of the project's manager.

### Changes

- **New `PackageManagerDetector`** (`packages/sdk/src/utils/PackageManagerDetector.ts`): walks up from the target directory to the fs root. Per level: `package.json` `packageManager` field first (explicit declaration beats lockfiles, so a project pinning `yarn@4` isn't hijacked by a stray `pnpm-lock.yaml`), then `pnpm-lock.yaml`/`pnpm-workspace.yaml` → pnpm, `yarn.lock` → yarn, `package-lock.json` → npm (a positive stop marker, so a nested npm project inside an outer pnpm monorepo keeps its own boundary). npm fallback at the root. Takes an injectable fs view for tests.
- **Distribute compose** detects from the distribute directory, composes manager-appropriate commands via a pure exported builder (`SdkDistributionComposer.commands`: `npm install --save(-dev)` / `pnpm add (-D)` / `yarn add (-D)`), and copies bundle files **write-if-missing** (`copyBundle` returns what it created; `${root}`/`${output}` placeholders are only patched in files that run created). User-authored `README.md`/`tsconfig.json`/`package.json` are never clobbered — same contract as the SDK output bundling after #1530.
- **`nestia dependencies`** keeps the explicit `--manager` override (now validated against `npm|pnpm|yarn`), defaults via detection from cwd, and pins: `@nestia/e2e@^<sdk version>` `@nestia/fetcher@^<sdk version>` `typia@<range @nestia/sdk declares>`, read at runtime from the SDK's own manifest (shared `SdkManifest` reader, also used by the composer). `catalog:`/`workspace:` ranges fall back to unpinned typia — those only exist inside this monorepo, never in the published package.
- **Removed `npx typia setup --manager npm`** instead of parameterizing it: **typia v13 deleted the `setup` subcommand** (verified in the installed CLI — only `generate` remains; anything else prints usage and exits −1), so any invocation would abort compose with an exception. The @next line wires the transform through the ttsc plugin pipeline, not ts-patch, so nothing replaces it.
- **`TtscExecutor`** messages are now manager-neutral: `Reinstall "ttsc" as a devDependency (e.g. pnpm add -D ttsc).`

### Tests

New `tests/test-distribute` workspace (plain-node, function-per-file `test_<snake_case>` suite; requires the **built** sdk lib by absolute path since the exports map blocks deep imports), registered in the CI matrix as `distribute`. 17 cases cover: every detection branch (packageManager field, pnpm-lock, pnpm-workspace, yarn.lock, package-lock stop marker, npm fallback via injected fs, walk-up past marker-less package.json, field-beats-lockfile, nested-npm-beats-ancestor-pnpm), the three managers' exact command strings for both distribute (incl. conditional MCP/tgrid lines and their omission when versions are unknown) and dependencies (pinned specs), the typia catalog fallback, and the write-if-missing copy against the real bundle assets (pre-placed README preserved, missing files created, returned list exact).

## Verification

- `pnpm --filter ./tests/test-distribute start` — 17/17 passed (Windows 11, local).
- `pnpm --filter ./tests/test-sdk start --only distribute` — 4/4 green. **Note:** these features are trivially green — `tests/test-sdk/start.js` short-circuits every feature whose name contains "distribute" before any generation (`if (name.includes("distribute")) return;`, present since the file's creation on this line), so compose has never executed in CI here.
- Compose exercised end-to-end in a scratch project with `child_process.execSync` stubbed: npm detected from `package-lock.json` two levels up, pre-placed README preserved, placeholders patched only in copied files, command order `rimraf → fetcher → typia → (tgrid)` with no `typia setup`.
- `nestia dependencies --manager bogus` halts with `Unsupported package manager "bogus". Choose one of (npm|pnpm|yarn).`
- Changed files formatted with the repo prettier config.

## The CI distribute-feature package-manager question

Asked to reason about whether pnpm detection would corrupt the repo workspace when the test-sdk distribute features run compose in CI. Findings:

1. **Compose does not run in CI at all**: `distribute:` configs exist only in the 4 `tests/test-sdk/features/distribute*` fixtures, and `start.js` returns before generation for all of them. So this PR cannot mutate the root lockfile during tests.
2. The hazard is real if that ever changes: an empirical check (pnpm 10) showed `pnpm add` in a nested directory whose `package.json` is **not** a workspace member still resolves the enclosing `pnpm-workspace.yaml` and **registers the directory as a new importer in the ROOT `pnpm-lock.yaml`**, also creating root `node_modules` churn. The feature dirs carry no manager marker of their own, so detection from `features/distribute/packages/api` would walk to the repo root and pick pnpm.
3. Chosen design: keep detection pure (nearest explicit marker wins; nested `package-lock.json` stops the walk) — correct for user projects — and leave the fixtures untouched since they never execute compose (also per the scope constraints on this PR). If the distribute features are ever re-enabled to run compose for real, the fixture dirs must first get their own boundary marker (a local `pnpm-workspace.yaml`, or `package.json` + `package-lock.json` to pin npm as before); the old npm behavior only left gitignored `package-lock.json`/`node_modules` strays, which is why this was never noticed.

## Known gaps

- Real installs against a live registry were not executed (command composition is stub-verified and unit-tested instead).
- Inside this monorepo the composer would still emit `typia@catalog:samchon` (pre-existing: the manifest is read at runtime and `catalog:` is only rewritten at publish); harmless because compose never runs here, and published installs get the real range.
- `--manager` values outside `npm|pnpm|yarn` (e.g. `bun`) now halt with guidance instead of blindly running `<value> install`; bun users previously "worked" by accident and now need one of the supported managers.
- Yarn berry is only covered at command-syntax level (`yarn add -D` is valid for both classic and berry); a lockfile-less PnP project without a `packageManager` field falls back to npm.

https://claude.ai/code/session_01GYUZ95W7jUpsnTgxBmamAH

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,6 +39,7 @@ jobs:
           - { name: transform-options, run: "pnpm --filter ./tests/test-transform-options start" }
           - { name: e2e, run: "pnpm --filter ./tests/test-e2e start" }
           - { name: benchmark, run: "pnpm --filter ./tests/test-benchmark start" }
+          - { name: distribute, run: "pnpm --filter ./tests/test-distribute start" }
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
--- a/packages/sdk/src/executable/internal/DependenciesInstaller.ts
+++ b/packages/sdk/src/executable/internal/DependenciesInstaller.ts
@@ -1,0 +1,54 @@
+import cp from "child_process";
+
+import { PackageManagerDetector } from "../../utils/PackageManagerDetector";
+import { SdkManifest } from "../../utils/SdkManifest";
+
+/** Composes and runs the `nestia dependencies` installation. */
+export namespace DependenciesInstaller {
+  /**
+   * Composes the exact install command for the given package manager.
+   *
+   * `@nestia/e2e` and `@nestia/fetcher` are pinned to the running SDK's own
+   * version line, and `typia` to the range `@nestia/sdk` itself declares, so a
+   * fresh `npx @nestia/sdk dependencies` cannot pull an incompatible major from
+   * the registry. Workspace-relative ranges (`catalog:`, `workspace:`) are not
+   * installable from the registry, so `typia` falls back to unpinned in that
+   * case — it only occurs inside this monorepo itself.
+   *
+   * Kept pure (inputs -> string[]) so unit tests can assert the command strings
+   * without executing any install.
+   */
+  export const compose = (props: {
+    manager: PackageManagerDetector.Manager;
+    version: string;
+    typia: string | undefined;
+  }): string[] => {
+    const specs: string[] = [
+      `@nestia/e2e@^${props.version}`,
+      `@nestia/fetcher@^${props.version}`,
+      props.typia !== undefined && installable(props.typia)
+        ? `typia@${props.typia}`
+        : "typia",
+    ];
+    const prefix: string =
+      props.manager === "npm" ? "npm install" : `${props.manager} add`;
+    return [`${prefix} ${specs.join(" ")}`];
+  };
+
+  export const install = (props: {
+    manager: PackageManagerDetector.Manager;
+  }): void => {
+    const manifest: SdkManifest.IManifest = SdkManifest.read();
+    for (const command of compose({
+      manager: props.manager,
+      version: manifest.version,
+      typia: manifest.dependencies.typia,
+    })) {
+      console.log(`\n$ ${command}`);
+      cp.execSync(command, { stdio: "inherit" });
+    }
+  };
+
+  const installable = (range: string): boolean =>
+    !range.startsWith("catalog:") && !range.startsWith("workspace:");
+}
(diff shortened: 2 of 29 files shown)


samchon:
Closed per maintainer decision — the premise (package-manager detection for distribute/dependencies) is not wanted.

PR #1541, opened 2026-07-18
feat(website): retheme the site to a NestJS red and white two-tone

samchon:
## Intent

The website shipped bare Nextra 4 defaults pinned to `defaultTheme: "dark"` with no toggle, so it read as a generic black docs template with nothing tying it to NestJS. typia and ttsc have each since moved to their own brand two-tone; this gives nestia the equivalent, built on the red NestJS already uses.

The mechanism is deliberately the same one those two repos settled on, so the three sites stay maintainable as siblings: `<Head color>` feeding Nextra's primary ramp, `forcedTheme: "light"`, a single `global.css` overriding Nextra's DOM classes, and marker-span scoping via `body:has()` so sidebar theming applies to docs routes only.

## The palette

Colors were taken off nestjs.com and docs.nestjs.com directly (Playwright, sampling computed styles) rather than eyeballed. Both paint their accents `#ea2845`.

That literal brand red measures **4.30:1 on white — below WCAG AA**, so it is not the anchor. It survives as `--nestia-red-vivid` for decoration that never carries text. Everything text-bearing resolves to `--nestia-red` `#b81e35` = `hsl(351, 72%, 42%)`, 6.40:1.

| Token | Value | Contrast on white |
|---|---|---|
| `--nestia-red` (Nextra anchor) | `#b81e35` | 6.40:1 |
| `--nestia-red-deep` | `#851425` | 9.88:1 |
| `--nestia-red-mid` | `#cc243d` | 5.41:1 |
| `--nestia-red-vivid` | `#ea2845` | 4.30:1 — decoration only |
| `--nestia-ink` | `#1f1417` | 17.95:1 |

The navbar band's three stops were each checked for white text rather than only the endpoints, so there is no unreadable midpoint wherever the gradient lands behind a link.

`--nestia-red` and the `<Head color>` HSL triple are the same color written twice, in CSS and JSX. Nothing enforces that; the comment in `global.css` says so explicitly.

## Scope

- `global.css` (new) — palette, navbar band, sidebar wash, callouts, landing full-bleed
- `layout.jsx` ×3 — `forcedTheme: "light"`, branded logo with a white plate, `<Head color>`
- `[[...mdxPath]]/page.jsx` — docs marker span
- Six landing sections + `PALETTE.ts` + `HomeSectionHeading.tsx` (new)
- `blog.css` — dead `.dark` rules removed, moved onto brand vars
- `site.webmanifest` — `theme_color` now brand red

Two judgement calls worth surfacing for review:

**Callouts.** Nextra paints `type="info"` blue and the docs use it 27 times against 3 warnings, so the most repeated element in the body was blue under a red brand — the one clash typia and ttsc never hit, since their brand *is* blue. They go warm-neutral with the accent on the left edge and icon. Filling them with brand red was rejected: at 27 occurrences it would wash the docs pink and make routine asides look like danger warnings. Chosen by the repository owner over an all-red variant and over leaving the blue.

**The comparison table inverts red-is-bad.** Red is the brand, so painting "Traditional NestJS" in brand red would read as an endorsement of the option the section argues against. The weaker column goes neutral; the recommended one takes the red.

**Code panels stay dark on the white page.** The highlighter is pinned to shiki's `github-dark`; re-tinting a whole syntax theme for a light surface is a much larger change than this one.

## Verification

`next dev` against an isolated copy of the site, driven with Playwright:

- `/`, `/docs/`, `/blog/` all return **200 with zero console errors**
- `html.light` on every route — dark mode is gone, not reskinned
- `--nextra-primary-hue: 351deg`, saturation `72%`, lightness `42%`
- body `rgb(255,255,255)` on `rgb(31,20,23)`
- navbar resolves to `linear-gradient(110deg, rgb(133,20,37), rgb(184,30,53) 58%, rgb(204,36,61))`
- docs marker present only on `/docs/`; landing class only on `/`
- `.nextra-callout` → bg `rgb(250,248,248)`, 3px left border `rgb(184,30,53)`
- blog card → bg `rgb(255,247,248)`, border `rgb(242,202,208)`
- landing swept for stray non-brand chrome; the only non-red hues remaining are shiki `github-dark` syntax tokens inside the code panels, which are intended

Screenshots reviewed at 1440px for all three routes plus a full-page landing pass.

## Known gaps

- **`/editor` was not exercised.** It needs a built `@nestia/editor` tarball, absent from this worktree; it was stubbed out of the verification copy. The route is untouched by this branch. The speculative `.nestia-editor-page` CSS I had copied from typia's playground pattern was removed rather than shipped unverified — nestia's editor already lays itself out as a full-viewport absolute overlay, so nothing carried the class.
- **`pnpm format` was not applied.** Running it rewrites ~250 files across `packages/` from pre-existing repo-wide prettier drift, unrelated to this change, and exits 2. Those edits were reverted so they do not enter this PR. Note `website/` is outside the root `format` script's globs, and files this branch never touches (e.g. `HomeCodeHighlight.tsx`) already fail `prettier --check` — so website formatting is pre-existing drift either way, and reformatting here would have been noise.
- Only the light scheme exists now; there is no dark variant to keep in sync, by design.
- `site.webmanifest` has a pre-existing unrelated bug: its icon `src` paths point at `/android-chrome-*.png` while the files live under `/favicon/`. Left alone as out of scope.

https://claude.ai/code/session_01DrosyEcDNyrpdqNjJs74LM

### DIFF
--- a/website/mdx-components-blog.jsx
+++ b/website/mdx-components-blog.jsx
@@ -39,7 +39,9 @@ export function useMDXComponents(components) {
       }
 
       return (
-        <div className={hasToc ? "nestia-blog-post-layout" : undefined}>
+        <div
+          className={hasToc ? "nestia-blog-post-layout" : "nestia-blog-post-solo"}
+        >
           <div className="nestia-blog-post-main">
             {metadata?.ogImage ? (
               <img
--- a/website/public/favicon/site.webmanifest
+++ b/website/public/favicon/site.webmanifest
@@ -1,1 +1,1 @@
-{"name":"","short_name":"","icons":[{"src":"/android-chrome-192x192.png","sizes":"192x192","type":"image/png"},{"src":"/android-chrome-512x512.png","sizes":"512x512","type":"image/png"}],"theme_color":"#ffffff","background_color":"#ffffff","display":"standalone"}
+{"name":"","short_name":"","icons":[{"src":"/android-chrome-192x192.png","sizes":"192x192","type":"image/png"},{"src":"/android-chrome-512x512.png","sizes":"512x512","type":"image/png"}],"theme_color":"#b81e35","background_color":"#ffffff","display":"standalone"}
--- a/website/src/app/(docs)/[[...mdxPath]]/page.jsx
+++ b/website/src/app/(docs)/[[...mdxPath]]/page.jsx
@@ -24,8 +24,16 @@ export default async function Page(props) {
   const result = await importPage(params.mdxPath);
   const { default: MDXContent, toc, metadata } = result;
 
+  // The sidebar wash is scoped to docs routes via `body:has(...)` in
+  // global.css, so the landing keeps a plain white gutter. This marker is
+  // what that selector keys off.
+  const isDocsPage = params.mdxPath?.[0] === "docs";
+
   return (
     <Wrapper toc={toc} metadata={metadata}>
+      {isDocsPage && (
+        <span className="nestia-docs-page-marker" aria-hidden="true" hidden />
+      )}
       <MDXContent {...props} params={params} />
     </Wrapper>
   );
--- a/website/src/app/(docs)/layout.jsx
+++ b/website/src/app/(docs)/layout.jsx
@@ -8,11 +8,7 @@ const footer = (
     <br />
     <br />
     Copyright 2022 - {new Date().getFullYear()}{" "}
-    <a
-      href="https://github.com/samchon"
-      target="_blank"
-      style={{ color: "initial" }}
-    >
+    <a href="https://github.com/samchon" target="_blank">
       <u>Samchon</u>
     </a>{" "}
     & Contributors
--- a/website/src/app/(docs)/layout.jsx
+++ b/website/src/app/(docs)/layout.jsx
@@ -25,23 +25,19 @@ export default async function DocsLayout(props) {
       navbar={
         <Navbar
           logo={
-            <>
-              <img
-                src="/favicon/android-chrome-192x192.png"
-                width={32}
-                height={32}
-              />
-              <span
-                style={{
-                  fontWeight: "bold",
-                  fontSize: "1.2rem",
-                  paddingLeft: 15,
-                  paddingRight: 10,
-                }}
-              >
+            <span className="nestia-site-logo">
+              <span className="nestia-site-logo-mark" aria-hidden="true">
+                <img
+                  src="/favicon/android-chrome-192x192.png"
+                  width={32}
+                  height={32}
+                  alt=""
+                />
+              </span>
+              <span style={{ fontSize: "1.2rem", paddingRight: 10 }}>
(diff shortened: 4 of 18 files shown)


samchon:
Follow-up pass: drove the running site with Playwright across desktop (1440) and mobile (390), including the surfaces the first pass never opened — the mobile drawer, the search field, tag pages, and an article page. Three things came out of it.

**1. Headings were broken on every hand-written blog route** (`377818b3`)

Not a retheme regression — this predates the branch. Nextra ships Tailwind Preflight, which resets headings to `font-size: inherit; font-weight: inherit`. MDX headings get their scale back from Nextra's component mapping, but the blog index, the tag pages, and the post title in `mdx-components-blog.jsx` are built in JSX and never pass through it.

All three rendered at **16px / weight 400**, identical to body copy. On an article this inverted the hierarchy outright — the post title came out *smaller* than the `## Preface` heading below it:

| | before | after |
|---|---|---|
| `h1` on `/blog/`, `/blog/tags/*`, post | 16px / 400 | 41.6px / 800 |
| sibling `h2` on a post | 30px | 30px |

It was invisible under the old black styling because everything else was low-contrast too. Also gave a TOC-less post a container — `hasToc` false dropped the grid wrapper along with every width and padding constraint, leaving the article full-bleed.

**2. The landing had no section rhythm** (`0bd49e02`)

Below the hero, five sections were white on white and ran together as one column. Now alternating: hero (red) → compilation (white) → features (wash) → comparison (white) → simulator (wash) → sponsors (white). Uses the existing `--nestia-wash`, so no new color — only rhythm. Cards inside the banded sections moved wash → white so they don't vanish into the band.

**3. The footer never left the old theme** (`16804e31`)

Its only link carried `color: initial` inline in both layouts — a dark-theme holdover — resolving to pure black on white, so it read as plain text. Removing that wasn't enough: it then inherited Nextra's footer gray `oklch(0.446 0.03 256.802)`, Tailwind gray-600 at hue 256 — a blue-tinted neutral, the last cool color left after the retheme.

Worth flagging: an earlier revision of that rule shipped a stray `*/`, which made the parser swallow the `footer` selector together with the loose comment text and silently drop the color rule. The link rule after it still applied, so it looked half-working rather than broken — the kind of thing that survives a glance. Both stylesheets are now parse-checked with postcss (global.css 33 rules / 73 decls, blog.css 36 / 126, no malformed selectors).

**Checked and found fine, no change needed**

- Mobile drawer — links ink, active item red-on-white, section headers `--nestia-red-deep`. The white link I first measured turned out to be the active item, which is correct.
- No horizontal overflow on any route at 1440 or 390.
- Console clean across all routes; the only message is React DevTools' info line.
- Search input on the red band reads correctly. The index itself is unavailable in dev — Pagefind indexes built HTML — so it was verified for styling only, not results.

Verification gaps from the original body still stand: `/editor` was not exercised, and `pnpm format` was not applied.

Issue #1552, opened 2026-07-21
@nestia/e2e: DynamicExecutor builds a malformed module specifier for a cross-drive location on Windows


samchon:
## Problem

On Windows, `DynamicExecutor` fails with `MODULE_NOT_FOUND` when the test
`location` sits on a different drive letter than the installed `@nestia/e2e`
package. The module specifier it builds is a malformed hybrid — a relative prefix
glued onto an absolute path:

```
Error: Cannot find module './C:/tmp/features/test_alpha.js'
Require stack:
- .../packages/e2e/lib/DynamicExecutor.js
```

The error names a path the user never wrote and points at `@nestia/e2e`'s own
`lib` directory as the resolution root, so it reads like a broken install rather
than a path-handling defect.

## Evidence

`packages/e2e/src/DynamicExecutor.ts`:

```ts
const specifier = (location: string): string => {
  const relative: string = NodePath.relative(
    __dirname,
    NodePath.resolve(location),
  )
    .split(NodePath.sep)
    .join("/");
  return `./${relative}`;
};
```

`path.relative` cannot express a relative path between two Windows drives, so it
returns the **absolute** target instead. Prefixing `./` then yields a specifier
that is neither relative nor absolute, and it resolves against `lib/` and fails.

```js
const p = require("path");
const lib = "D:/.../packages/e2e/lib";

p.relative(lib, p.resolve("C:/tmp/features"));
// -> "C:/tmp/features"          (absolute — different drive)
// specifier -> "./C:/tmp/features"        BROKEN

p.relative(lib, p.resolve("D:/.../.wiki/scratch/features"));
// -> "../../../.wiki/scratch/features"   (genuinely relative — same drive)
// specifier -> "./../../../.wiki/scratch/features"   works
```

Observed end-to-end at `429099bf` against the built package: with `location` on
`C:` and `@nestia/e2e` on `D:`, `DynamicExecutor.validate` throws
`MODULE_NOT_FOUND` for `./C:/Users/.../test_alpha.js`. Moving the identical
fixture directory to the same drive as the package makes the same call succeed and
discover the test — that is the one-axis control, and it isolates the drive
difference from any fixture or option problem.

Independent oracle: `DynamicExecutor` must be able to import any readable
directory the caller names in `location`. `location` is documented only as
"Location of the test functions", with no same-volume restriction, and the value
is passed through `NodePath.resolve`, which explicitly accepts absolute paths.

[...]

Issue #1551, opened 2026-07-21
packages/*/native: the superseded text-splice transform path is unreachable but still shipped and compiled


samchon:
## Problem

The native transform was reworked from a text-splice pipeline to an
AST/emit-context pipeline, but the predecessor implementation was never removed.
It is unreachable from every entry point, and it is still published and compiled.

Two consequences go beyond tidiness:

1. **A registration seam exists with no consumer.** `@nestia/sdk` registers two
   contributor collectors that nothing ever calls, so the registry advertises
   three extension points while only one is live. A future contributor that picks
   the wrong one gets a transform that silently never runs.
2. **A green test guards a contract that no longer exists.**
   `packages/core/test/exports_apply_source_rewrites_test.go` asserts that
   `NewSourceRewrite` / `ApplySourceRewrites` are "the only way another Go module
   (the `@nestia/sdk` plugin) can produce one" — a seam no executed path uses. The
   test passes and proves nothing about shipped behaviour.

`packages/core/package.json` ships `native/cmd`, `native/plugin`, and
`native/transform`, and ttsc compiles that source on the consumer's machine on
first use, so the dead tree is type-checked and SSA-built in every consumer
project's cold plugin build and redistributed in every tarball.

### Which path is live

`runBuild` (`packages/core/native/transform/build.go`) and `runTransform`
(`transform.go`) both assemble
`[typiaTransform, coreTransform, ...contributorTransforms]` and hand them to
`prog.EmitWithPluginTransformers` / the printer. Their own comments state the
predecessor is gone:

> AST-integration emit: typia's per-file transformer and @nestia/core's own
> per-file transformer both run inside tsgo's emit pipeline (sharing the
> EmitContext) … **No text-splice RewriteSet, no cleanup pass.**

The text-splice implementation those comments describe as absent is still in the
tree.

### Reachability

Roots: `main` → `transform.Run`, every `init()`, everything referenced from
package-level declarations, and the exported seams (`NewSourceRewrite`,
`ApplySourceRewrites`, `SourceFileText`, `NodeName`, `NodeText`, `ParsePlan`,
`NestiaCoreMethodReturnType`, `WriteTypiaTransformDiagnostics`, `EmitTransform`,
`ApplyProgram`). Comments stripped before matching, so a mention in prose cannot
keep a symbol alive.

Result at `429099bf`: **79 of 373 top-level functions are unreachable**, in these
files:

| File | dead funcs | total lines | fully dead? |
[...]

Issue #1550, opened 2026-07-21
`pnpm format` aborts on a nonexistent `internals/` glob and never formats `tests/**/*.ts`


samchon:
## Problem

`pnpm format` never formats `tests/**/*.ts`. The root script chains three Prettier
invocations with `&&`, and the middle one targets a directory that does not exist:

```json
"format": "prettier --write packages/**/*.ts && prettier --write internals/**/*.ts && prettier --write tests/**/*.ts"
```

There is no `internals/` directory in this repository. Prettier exits **2** when a
pattern matches nothing, so the `&&` chain aborts and the third invocation never
runs.

The development skill requires `pnpm format` before every ordinary commit, so every
contributor and agent that follows the documented workflow believes the test
workspaces are formatted. They are not, and the gap has grown to over a thousand
files.

## Evidence

```
$ ls internals
ls: cannot access 'internals': No such file or directory

$ npx prettier --check "internals/**/*.ts"
Checking formatting...
[error] No files matching the pattern were found: "internals/**/*.ts".
All matched files use Prettier code style!
$ echo $?
2
```

Exit code 2 with `&&` chaining is the whole mechanism: the first invocation
succeeds, the second exits non-zero, and the shell stops.

Impact measured at `429099bf` with the repository's own `prettier.config.js` and
`.prettierignore`:

```
$ npx prettier --list-different "tests/**/*.ts" | wc -l
1589      # every tracked TypeScript file under tests/
1218      # excluding the generated */src/api/** trees
```

1218 hand-written, tracked files violate the repository's own Prettier
configuration — across `test-benchmark`, `test-cli`, `test-e2e`, `test-editor`,
`test-migrate`, `test-sdk`, and `test-transform-options`.

Independent oracle: the repository's intent is already recorded, not inferred.
`.agents/skills/project/SKILL.md` states that `pnpm format` "does not touch Go,
Markdown, MDX, the website, or the benchmark workspace", explains this exact
abort, and instructs contributors to run `npx prettier --write "tests/**/*.ts"`
manually **"until the root script is repaired."** The script is the defect; the
target list is correct.

### Ownership

Root `package.json` of this repository.

## Consequence surface

- Every ordinary commit made through the documented workflow since the `internals/` directory disappeared.
- All seven `tests/test-*` workspaces.
- No package build, test, or CI lane fails today, which is precisely why it went unnoticed: nothing checks formatting.
[...]

Issue #1549, opened 2026-07-21
@nestia/e2e: DynamicExecutor discovers zero tests for any extension longer than two characters


samchon:
## Problem

`DynamicExecutor` silently discovers **zero** test functions when the `extension`
option is anything other than a two-character extension. `extension` is a
documented public option with `@default js`, so `"mjs"`, `"cjs"`, and `"tsx"` are
all legitimate values — and all three make the executor find nothing.

The failure mode is the dangerous one: `DynamicExecutor.validate` returns a
well-formed report with an empty `executions` array and no error, so a suite that
does not separately assert "at least one test ran" reports success while having
run nothing.

This exact failure has already shipped once. `.agents/skills/development/SKILL.md`
records that "a hardcoded `js` extension once made a suite pass vacuously", and the
response was to add a zero-test guard to `tests/test-e2e` and `tests/test-editor`.
The guard catches the symptom in two in-repo suites; the predicate that produces it
is unchanged, and no downstream consumer of the published package has that guard.

## Evidence

`packages/e2e/src/DynamicExecutor.ts`, inside `iterate`:

```ts
if (stats.isDirectory() === true) {
  await visitor(location);
  continue;
} else if (file.substr(-3) !== `.${props.extension}`) continue;
```

`substr(-3)` takes the last three characters, which equals `"." + extension` only
when `extension` is exactly two characters long. For a three-character extension
the comparison is `"mjs" !== ".mjs"` and can never be true.

The correct shape already exists in the sibling package —
`packages/benchmark/src/DynamicBenchmarker.ts`:

```ts
file.endsWith(`.${ctx.props.extension ?? "js"}`) &&
```

so the two published discovery paths disagree about the same public option.

`String.prototype.substr` is additionally an Annex-B legacy function.

### Reproduction

Two fixture files in one directory, differing only in extension:

```
features/test_alpha.js    ->  exports.test_alpha = async () => "js";
features/test_beta.cjs    ->  exports.test_beta  = async () => "cjs";
```

Against the built package (`pnpm --filter ./packages/e2e build`):

```js
const { DynamicExecutor } = require("./packages/e2e/lib/index.js");
for (const ext of ["js", "cjs"]) {
  const report = await DynamicExecutor.validate({
    prefix: "test",
    location: "<abs>/features",
    parameters: () => [],
    extension: ext,
  });
  console.log(ext, report.executions.length, report.executions.map((e) => e.name));
}
```

Observed at `429099bf`:

```
[...]

PR #1553, opened 2026-07-21
fix: issue campaign cycle 1

samchon:
## Intent

This pull request owns the **complete accepted cycle** of the 2026-07 solo issue
campaign. Every implementation-ready issue discovered in the campaign's first
round lands here as one integrated unit; the issue DAG controls edit order inside
the branch, not pull-request count.

Opened as an empty claim before any code exists, so overlapping work is visible
early.

## Cycle scope

| Issue | Package | Summary |
| --- | --- | --- |
| #1547 | `@nestia/fetcher` | a trailing slash in `connection.host` produces a double-slash request URL |
| #1548 | `@nestia/e2e` | `TestValidator.equals` / `notEquals` throw `TypeError` when the actual value is `null` |
| #1549 | `@nestia/e2e` | `DynamicExecutor` discovers zero tests for any extension longer than two characters |
| #1550 | repo root | `pnpm format` aborts on a nonexistent `internals/` glob and never formats `tests/**/*.ts` |
| #1551 | `packages/*/native` | the superseded text-splice transform path is unreachable but still shipped and compiled |
| #1552 | `@nestia/e2e` | `DynamicExecutor` builds a malformed module specifier for a cross-drive `location` on Windows |

One further candidate (typia plugin options read by regex over raw tsconfig text
instead of the resolved `--plugins-json` payload) is **not** in this cycle. It is
recorded as deferred pending a real `ttsc` compile that this campaign round could
not execute; it will be published only once reproduced.

### Dependencies

- #1549 and #1552 touch adjacent lines of the same function (`DynamicExecutor.iterate` / `specifier`) but are independent causes with separate acceptance criteria.
- #1550 changes which files the formatter rewrites, so it is ordered so the mechanical reformat stays separable from every behavioural diff.
- #1547, #1548, #1551 are independent of everything else.

## Verification plan

- Narrow, per issue: `pnpm --filter ./tests/test-e2e start`, `pnpm --filter ./tests/test-sdk start -- --only cli`, `pnpm --filter @nestia/core test:go`, `pnpm --filter @nestia/sdk test:go`.
- Broad: `pnpm build`, `pnpm test:go`, `pnpm test`, `pnpm format`.
- `packages/*/native` is integration-sensitive, so #1551 additionally gets the frozen integration-command set against the prospective merge result before merge.
- Regression sensitivity is proved by mutation: each product fix is reverted in a disposable checkout with its tests retained, and the expected assertions must fail while adjacent controls still pass.

## Honest status at open

- Nothing is implemented yet; this is the claim commit.
- Local toolchain is Node 22.21.0 against the repository's Node 24.x contract, so
  the pull request's own CI is the authoritative lane for every suite. Local runs
  are supporting evidence, not a substitute.
- Package versions, release tags, and publication state are frozen for the whole
  campaign. This pull request changes no `version` field.


samchon:
## Implementation ledger

Posted as one ledger rather than per commit — the batching is on me, and the
mapping below is what a reader who does not read the diff needs.

| Commit | Lands | Resolves |
| --- | --- | --- |
| `55f4d944` | `fetcher`: host/path join owns both boundaries in one internal helper; `NestiaSimulator` composes its error-payload path through the same helper | #1547 |
| `284215a3` | `e2e`: null pairs settled before the array/object branches in the deep comparator | #1548 |
| `09438b54` | `e2e`: whole-suffix extension match; specifier falls back when no relative expression exists | #1549, #1552 |
| `8d385fe0` | `sdk`: value-position bound in `getFileArgument`; new `cli-argument-diagnostics` harness feature | #1554 |
| `893e398e` | `native`: superseded text-splice path removed (126 functions); stale seam test deleted; `matchNativeParen` moved to its only consumer | #1551 |
| `d4e2cd52` | root: one formatter invocation, unparseable template ignored, `**/api/functional` anchored, three skill docs corrected | #1550 |
| `4dfd456d` | mechanical `pnpm format` result, 62 files, no behavioural diff | — |
| `4b6bd0f9` | en-us spelling in a test doc comment (spell-check lane) | — |
| `dcb0e02c` | `test-transform-options`: launch ttsc through node, report spawn failures instead of crashing | #1555 |
| `db8c6a1e` | `native`: dead types and vars the removed seams left behind, plus gofmt | — |

## Two corrections to a published issue

#1550's body was wrong twice and both are recorded on that thread rather than
edited away:

1. The impact figures (1589 / 1218 files) were a **line-ending artifact** — this
   checkout has `core.autocrlf=true` and `prettier.config.js` leaves `endOfLine`
   at `"lf"`, so Prettier called every file different. Re-measured with
   `--end-of-line auto`: **40** tracked files under `tests/`, not 1218.
2. The published root cause was **not the first abort**. `packages/**/*.ts` also
   exits 2, because `packages/sdk/assets/bundle/e2e/index.ts` is a template whose
   `${input}` placeholder sits at statement position and Prettier cannot parse it.
   The `internals/` glob is real but latent behind it. Removing only the
   `internals/` invocation would have left `pnpm format` still failing.

## One issue discovered during implementation

#1555. `tests/test-transform-options` spawned the `ttsc` `.cmd` shim without a
shell, which Node refuses, so the suite could not run on Windows at all — and its
`fail: true` cases passed **vacuously**, because an EINVAL spawn satisfies
`status !== 0` just as well as a rejected compile. Published, then fixed here,
because it also blocks local verification of the native change on Windows.

## Verification

Executed locally at head, Windows 11 / Node 22.21.0 / Go 1.26.4:

| Command | Result |
| --- | --- |
| `pnpm build` | pass (after the native removal) |
| `pnpm test:go` | pass — core 7.8s, sdk 14.8s |
| `go build ./...` + `go vet ./...`, both native modules | pass |
| `pnpm --filter ./tests/test-e2e start` | pass, 15 tests |
| `pnpm --filter ./tests/test-sdk start -- --only cli-argument-diagnostics` | pass |
| `pnpm --filter ./tests/test-transform-options start` | pass — could not run at all before #1555 |
| `pnpm format` | exit 0 (was exit 2) |
| reachability over both native modules | 0 unreachable functions (was 79) |

### Mutation sensitivity

Each product fix was reverted in place with its tests retained, the narrow suite
re-run, and the worktree restored to a zero diff. Every one failed at exactly the
intended assertion while the preceding controls passed:

| Reverted | Failing assertion | Controls before it |
| --- | --- | --- |
| #1547 join | `Bug on "https://example.com/" + "/health"` | 3 tests pass |
| #1548 null branch | `TypeError: Cannot read properties of null (reading 'name')` | 7 tests pass |
| #1549 predicate | `Bug on three-character extension: found different values` | 2 tests pass |
| #1552 specifier | `Cannot find module './C:/Users/.../test_remote.js'` | 1 test passes |
| #1554 guard | `Failed test-sdk features: cli-argument-diagnostics` | n/a (single feature) |

#1551 does not take a mutation: a dangling reference to a removed symbol is a hard
Go compile error, so `go build`, `go vet`, `pnpm test:go`, `pnpm build`, and the
`test-transform-options` suite that inspects the exact arguments the transform
bakes into decorator calls are its oracle. #1550's oracle is the exit code itself.

## Honest gaps

- Local Node is 22.21.0 against the repository's 24.x contract, so this pull request's CI is the authoritative lane. `pnpm test` was not run end to end locally; the suites above were run individually.
- `tests/test-migrate`, `test-benchmark`, `test-cli`, `test-editor` were not run locally and rest on CI.
- CI is `ubuntu-latest` only, so the Windows half of #1552 and #1555 is proved by local execution, not by a lane.

samchon:
Correction to the ledger above: the last three rows carry placeholder hashes I
wrote before those commits existed. The real ones are

| Ledger row | Wrong | Actual |
| --- | --- | --- |
| en-us spelling | `4b6bd0f9` | `69410902` |
| test-transform-options launch fix | `dcb0e02c` | `c856ed3f` |
| dead types + gofmt | `db8c6a1e` | `a77a6f9b` |

Every other row is correct. Nothing about what landed changes.

samchon:
## Ledger, continued

| Commit | Lands | Resolves |
| --- | --- | --- |
| `07c4e36f` | separator boundaries scanned instead of regex-matched; fixture assigns through a quoted key | — (CodeQL) |
| `ea258b90` | 14 comments that still explained the emit path by contrast with the removed one | — |
| `6a745a03` | cli-argument-diagnostics reports a launch failure as itself | — |

### CodeQL found two real defects in this branch's new code

Both fixed in `07c4e36f`, neither pre-existing:

- **high, `js/polynomial-redos`** — `join_host_and_path` matched the trailing separator with `/\/+$/`. `host` is caller-supplied and that pattern backtracks quadratically over a run of separators, so a pathological host would cost that on every request. Both boundaries are now index scans, which is also cheaper in the ordinary case.
- **medium, `js/bad-code-sanitization`** — the extension-width fixture interpolated a name into an identifier position. It now assigns through a `JSON.stringify`d key.

### Self-Review found what the tooling could not

- Function-level reachability enumerates **funcs**, so unreferenced types and package-level vars survived the first removal pass: `BuildOutputRewriter`, both collector function types, both collector slices, `SourceRewrite`, `nestiaTypiaImportRoot`. Removed in `a77a6f9b`.
- Fourteen comments still described the deleted text-splice path as a live alternative — one named `nestiaCoreParameterArguments`, a function this branch deletes, as a current caller. Corrected in `ea258b90`; the nil guards themselves stay, since nothing proves a future caller cannot pass nil.
- My own new `cli-argument-diagnostics` feature resolved a spawn error into the captured-output string, so a CLI that never started would have been reported as a CLI that printed the wrong diagnostic — exactly the confusion #1555 fixes one workspace over. Corrected in `6a745a03`.

## One issue found while this settled, deliberately not folded in

**#1556** — typia transform options declared in a base tsconfig are silently
dropped through `extends`. `readTypiaPluginOptions` regex-scans one file's raw
text instead of reading the resolved `--plugins-json` payload ttsc already hands
the binary.

Differential compile at `429099bf`, same source, three configs:

| config | emitted `feature.js` |
| --- | --- |
| no `numeric` | 3184 bytes |
| `"numeric": true` inline | 3306 bytes |
| the same option through `extends` | **3184 — byte-identical to no option** |

The `!Number.isNaN(...)` guard is simply absent from every generated validator.

It is **not** in this pull request. The reproduction needed a real `ttsc` compile
and landed after this head's evidence had settled; a native change now would
restart CI, Self-Review, and the integration gate on an otherwise-final head. It
is cycle 2's first implementation-ready issue.

## Integration gate

`origin/master` is still `429099bf`, the claim base, so the prospective merge
result **is** this head — the gate reduces to the canonical command set on
`6a745a03`, which this run executes: root build (`Ubuntu`), all eight test lanes,
packaging and a clean consumer (`deploy` → `package:tgz` + website install/build),
CodeQL, and spell-check.

Issue #1590, opened 2026-07-22
test-sdk: give the runtime Swagger error regression its own test case file


samchon:
## Problem

The regression test added for #1577 exports two independently discovered test cases from `test_runtime_swagger.ts`. This violates the repository test contract and removes the new error case's file-level identity and required explanatory doc comment.

## Evidence

The development contract predates #1578 and requires one new or changed test case per matching file. `tests/test-sdk/features/app/src/test/features/test_runtime_swagger.ts` now exports both `test_runtime_swagger` and `test_runtime_swagger_errors`. The project-shaped SDK harness executes every prefixed export discovered under `src/test/features`, so both cases run; this issue is a test-oracle structure violation, not a skipped-test claim.

## Consequence surface

The ordinary runtime Swagger document assertion and the manually injected route-analysis error assertion are independent behaviors. Splitting them must preserve both execution paths and avoid changing generated SDK fixture output.

## Approach

Give the error assertion a matching independently discovered file and a three-part case doc comment. Keep the ordinary document assertion in its own file.

## Acceptance and verification

The Node 24 SDK CI feature must discover and execute both file-level cases independently. The error case must retain the route-analysis rejection assertion; the ordinary case must retain the valid Swagger document assertion.

## Coordination

This follows #1578 and is owned by `tests/test-sdk`. It does not reopen #1577, whose product fix remains merged.

Issue #1589, opened 2026-07-22
@nestia/sdk native: classify TypeGuardError by typia provenance


samchon:
## Problem

A user-defined type named `TypeGuardError` is treated as typia's validation exception during native SDK metadata generation. Its declared exception schema is replaced with the synthetic typia error schema.

## Evidence

`nestiaSDKIsTypeGuardError` returns true from displayed type spelling (`TypeGuardError` or `TypeGuardError<...>`) before consulting import provenance or checker symbol ownership. `nestiaSDKSchemaPipe` then bypasses ordinary analysis and injects the hard-coded `TypeGuardErrorany` schema.

## Consequence surface

A local or non-typia imported same-named exception in `@TypedException` can change Swagger components, required fields, and generated SDK exception typing. Typia's real error type must keep its special schema. Aliases and re-exports require the same provenance-sensitive decision.

## Approach

Determine special typia-error handling from resolved checker/import ownership rather than raw spelling. Keep normal metadata analysis for every non-typia same-named type.

## Acceptance and verification

In Node 24 SDK CI, generate Swagger and SDK from a controller using a local or external non-typia `TypeGuardError` in `@TypedException`; assert its declared fields survive. Keep a typia `TypeGuardError` positive control and alias/re-export coverage where the checker can resolve provenance.

## Coordination

Historical issue #489 and PR #537 concern runtime conversion of typia errors to HTTP 400 responses. They do not modify this SDK native-transform classification path.

Issue #1588, opened 2026-07-22
@nestia/sdk native: preserve named-import aliases in generated SDK metadata


samchon:
## Problem

Default non-clone SDK generation can emit an import for a controller's local alias rather than the source module's exported name. The generated SDK then fails type checking for `import { User as Account } from ./models`.

## Evidence

`nestiaSDKAnalyzeImports` records only `ImportSpecifier.Name()`, the local binding `Account`. `IReflectImport` carries `elements: string[]`, so the native-to-generator ABI loses `User` before `ImportDictionary` emits declarations. The resulting metadata can request `Account` from a module that exports only `User`.

## Consequence surface

Named imports with aliases in non-clone HTTP SDK generation are affected for parameters, returns, and exceptions. Default and namespace imports are separate forms. Clone mode replaces ordinary controller imports with generated structures and needs explicit compatibility review.

## Approach

Carry enough imported-name and local-binding information through native metadata for the SDK generator to recreate a resolvable TypeScript import. Preserve the existing metadata behavior for unaliased, default, and namespace imports.

## Acceptance and verification

In Node 24 SDK CI, use a controller whose parameter and return DTO are imported under a different local alias. Generate the default SDK and compile it. Keep an unaliased control, plus applicable clone/WebSocket controls, and inspect the emitted import binding rather than accepting a compile-only coincidence.

## Coordination

Closed PR #1339 fixed an earlier TypeScript-side path/type alias case in `DtoAnalyzer`; it predates and does not modify the Go contributor import ABI.

Issue #1587, opened 2026-07-22
@nestia/sdk native: do not suppress same-named controller methods in namespaces


samchon:
## Problem

The SDK native metadata transform can skip a decorated controller method when two namespace-qualified classes in the same source file share the same local class and method names.

## Evidence

`collectNestiaSDKSites` creates a per-file visited map keyed only by `ClassName.MethodName`. `nestiaSDKParentClassName` returns the nearest local class name, so both `namespace First { class Foo { get() {} } }` and `namespace Second { class Foo { get() {} } }` produce `Foo.get`. The second method is skipped before metadata extraction and decorator injection even though both declarations are distinct.

## Consequence surface

Core decorators can still compile and runtime reflection can discover both controllers while one lacks SDK `OperationMetadata`. SDK, Swagger, and generated e2e output can therefore be incomplete or inconsistent. Controllers with distinct local class names remain a control case.

## Approach

Identify a metadata site by its distinct declaration identity within the source file, including lexical qualification where relevant, while preserving one injection per actual decorated method.

## Acceptance and verification

In Node 24 CI, compile/register two namespace-qualified controllers with identical local class and method names but different paths. Generate SDK and Swagger, then assert both operations and their metadata survive. Keep a distinct-class-name control and verify no duplicate injection for one declaration.

## Coordination

Owned by `packages/sdk/native`; no historical native-transform issue resolves this visited-key collision.

PR #1591, opened 2026-07-22
fix(campaign): resolve cycle 4 repository issues

samchon:
## Claim

This draft owns the complete implementation-ready cycle 4 issue set:

- #1581
- #1582
- #1583
- #1584
- #1585
- #1586
- #1587
- #1588
- #1589
- #1590

The internal order is recorded in the campaign ledger. No issue-closing keyword is present in this claim because no implementation has landed yet.

## Scope

The cycle covers valid empty OpenAPI migration, editor default composition, WebSocket diagnostics, root benchmark portability/CI, SDK distribution state, generated-output paths, native SDK metadata identity/import/provenance, and the runtime Swagger regression test layout.

## Verification

Pending implementation. Per the user's explicit directive, no local build, test, generator, or formatter has run. The acceptance gate is GitHub Actions CI under Node 24 plus a formal solo Self-Review for the final immutable head.

## Deferred work

After this campaign completes through its required empty post-merge rediscovery round, a separate campaign will reduce `test-sdk` duration using evidence from CI timings while preserving end-to-end transform and generated-artifact coverage.


samchon (commented):
Commit `813c7090` is the implementation-free cycle claim. It creates no source, test, workflow, or generated-artifact change and resolves no issue. The complete owned set is #1581 through #1590; verification remains pending GitHub Actions CI under the user’s no-local-execution directive.

samchon (commented):
Self-review for e6fe31357fc257d098dcc07e3cb917f6cb930d5e: verified the worker bound never becomes negative; the legacy CPU probe is delayed and falls back through Node; all three autocannon callers share it; and the CI path includes benchmark sources plus Ubuntu/Windows build-and-test jobs. No local build, test, generator, or formatter was run; GitHub Actions is the validation authority.

samchon (commented):
Follow-up self-review for 1c78415ad3f836c9ed0aa84d50f08d00f32171a3: the mandatory early-warning pass found the remaining BenchmarkServer import, and this follow-up removes it. The CI smoke now imports that runner bootstrap, exercises the real CPU probe under both OSes, and does not execute a benchmark workload. No local build, test, generator, or formatter was run.

samchon (commented):
Self-review for 8e5a8d7ea9866b48f520f2673e3a2bde358e168f: a valid empty OpenAPI Paths Object now produces a no-op 	est/start.ts instead of dereferencing a missing route. Nonempty documents retain the sample route path; the new CI-only migrate feature asserts the empty starter has no connection or functional imports. No local build, test, generator, or formatter was run.

samchon (commented):
Follow-up self-review for 7fd6e530b157745f86b8f3d6464305d2957a016d: the early-warning review correctly found that the empty-path case needed generated-output compilation, not only text assertions. test-migrate now archives the empty SDK and runs both its root and test tsconfig compiler steps exclusively in GitHub Actions. No local build, test, generator, or formatter was run.

samchon (commented):
Self-review for d92d2eee3d9d59e3c5c68ef260aeaa2f2bcd3155: corrected the uploader's default slug to @ORGANIZATION/PROJECT, centralised it in the shared internal value used by the uploader, and added a test-editor check against the built artifact plus its composed SDK manifest. No local build, test, generator, or formatter was run.

samchon (commented):
Follow-up self-review for 08cfbb5ad69b1cabaa2aa9d47409874f852ce8b5: the required empty-SDK compile scenario exposed a missing functional root. The migration writer now emits src/functional/index.ts even with zero routes, and the existing CI-only archive/root/test compilation path verifies it. No local build, test, generator, or formatter was run.

samchon (commented):
Follow-up self-review for 358a40f075de3e9bde674f4169014d714e247a27: the early-warning review found that reading a standalone constant did not prove the uploader used it. The test now server-renders the built Uploader, extracts its initial package input, and uses that observed value for SDK manifest propagation. No local build, test, generator, or formatter was run.

samchon (commented):
Final self-review for 08cfbb5ad69b1cabaa2aa9d47409874f852ce8b5: mandatory early-warning PASS. The zero-route root is now always emitted for SDK and Nest paths; existing fixture modes preserve the same root entry, and the new empty SDK CI case archives then compiles root plus test projects. No local execution was run.

samchon (commented):
Final self-review for 358a40f075de3e9bde674f4169014d714e247a27: mandatory early-warning PASS. The built Uploader SSR is safe for this component path, observes the initial input value, and test-editor builds that artifact before the test; the observation drives the SDK manifest assertion. No local execution was run.

samchon (commented):
Self-review for 7f525a17c67c9b88d24750cc89f97c1b6bd3c278: all invalid WebSocket parameter branches now reject into project diagnostics rather than entering the typed-route array. A dedicated CLI fixture bypasses tgrid's own generic check, captures the real CLI failure, and requires the public Acceptor arity diagnostic. No local build, test, generator, or formatter was run.

samchon (commented):
Follow-up self-review for d530e4f6aa318d963bc6210de795335977e46d36: the mandatory early-warning pass found a potential stderr/stdout drain race. Captured CLI diagnostics now finalize at child close rather than exit, so the exact public error assertion waits for both pipes. No local execution was run.

samchon (commented):
Final self-review for d530e4f6aa318d963bc6210de795335977e46d36: mandatory early-warning PASS. The child close event is after its stdout/stderr pipes close, preserving the exact nonzero CLI diagnostic while keeping success/signal failures explicit. No local execution was run.

samchon (commented):
Self-review for 78ca51156b2ac3c95dc3ee95e8ad93c39bd143a4: distribution composition now scopes every post-chdir operation in try/finally. The CI-only regression runs in a child process, stubs the first npm call after chdir, asserts restoration before returning, and removes its OS temp directory—without perturbing parallel SDK features. No local execution was run.

samchon (commented):
Final self-review for 78ca51156b2ac3c95dc3ee95e8ad93c39bd143a4: mandatory early-warning PASS. try/finally covers every failure after chdir; the regression uses a unique OS temp root in its own child, stubs the first install, verifies the original cwd, restores the stub, and awaits cleanup. No local execution was run.

samchon (commented):
Self-review for 873f99935a4d976af1af22f36a0d7d1059367583: split the two DynamicExecutor exports into one file each. The normal endpoint and invalid-metadata composer cases retain their behavior, now each has a Verifies/Why/numbered-steps contract, and discovery continues over the same features directory. No local execution was run.

samchon (commented):
Final self-review for 873f99935a4d976af1af22f36a0d7d1059367583: mandatory early-warning PASS. DynamicExecutor discovers both single-export test files recursively; the invalid-metadata fixture remains isolated and closes its app in finally. No local execution was run.

samchon (commented):
Self-review for c9c354f0f824c3c8c935cae9f138f3c589c49cce: fixed Swagger's parent calculation and made Swagger, SDK, E2E, SDK-file, and distribution roots create missing ancestors recursively. The new SDK CI fixture uses separate nested SDK/E2E/Swagger roots, while the isolated distribution regression now asserts a nested distribution root too. No local execution was run.

samchon (commented):
Self-review for eca45a04: nested-output regression now removes the fixture's generated tree before each run and explicitly requires concrete SDK, Swagger, and E2E artifacts. Mandatory early-warning review passed; no local build, test, generator, or formatter was run.

samchon (commented):
Self-review for caab0079: namespaced controller aliases now make both collision witnesses discoverable while preserving identical class/method spellings. The native collector uses per-file AST method positions, and the Swagger assertion verifies both routes. Mandatory early-warning review passed; no local build, test, generator, or formatter was run.

samchon (commented):
Self-review for 770c95a1: named-import aliases now preserve local and exported names through HTTP and WebSocket clone generation; TypeGuardError classification resolves aliases and verifies the resolved package root through the compiler VFS. Regression fixtures cover the negative/positive provenance cases and cloned WebSocket import path. Mandatory early-warning review passed. No local build, test, generator, or formatter was run.

samchon (commented):
Full self-review at 770c95a1: inspected the complete diff against f1f35cfd, including benchmark CI portability, empty migrate output, editor default propagation, WebSocket diagnostics, nested outputs/cwd restoration, native collector/import/provenance changes, and all regression harnesses. No further sound issue found. Working tree contains only the pre-existing untracked packages/core/test/method_return_type_complex_promise_test.go; no local build, test, generator, or formatter was run.

samchon (commented):
CI failure self-review at d48db859: the failing provenance fixture exposed a general SDK bug—exception imports were emitted even when propagation mode does not reference exception types. Reviewed the propagation response path, non-propagated SDK/e2e paths, and clone referencer. Imports are now retained only for propagate=true; the fixture still proves typia alias positive and local same-name negative Swagger schemas. Independent read-only early-warning review: PASS. No local build, test, generator, or formatter was run.

samchon (commented):
CI failure self-review at 17d6a34e: four jobs failed in package build with the same verified TS6196 on the legacy TypedException<T>(status, description) overload. T is intentionally consumed by the native SDK transformer from the decorator call-site; it has no TypeScript-level occurrence in that legacy signature. Added a self-invalidating @ts-expect-error directly on the affected overload, preserving all parameter/return and runtime behavior. Independent read-only early-warning review: PASS. No local build, test, generator, or formatter was run.

samchon (commented):
CI failure self-review at ed4dcffd: the import fix initially crossed a deliberate reflection boundary by reading config from an Omit<INestiaProject, config> context. NestiaSwaggerComposer legitimately uses the same controller analyzer without config. The HTTP analyzer now accepts config only optionally: normal SDK generation preserves the same runtime project reference and retains exception imports for propagate=true; config-less Swagger reflection safely omits them. Clone exception structures remain separately reconstructed. Independent read-only early-warning review: PASS. No local build, test, generator, or formatter was run.

samchon (commented):
CI failure self-review at a01e3b3b: the prior suppression was configuration-sensitive—some test compilers do not emit TS6133, making @ts-expect-error itself TS2578. The legacy overload now mentions T only through Pick<TypedException.IProps<T>, never>: structurally empty, runtime-free, and preserving the MethodDecorator contract while keeping the transformer-visible generic instantiation. No local build, test, generator, or formatter was run.

samchon (commented):
Independent early-warning review for a01e3b3b: PASS. The empty Pick intersection is structurally MethodDecorator & {}, preserves deprecated generic decorator compatibility, does not alter emitted runtime/reflection metadata, and the native transformer reads call AST type arguments directly. No local build, test, generator, or formatter was run.

Issue #1598, opened 2026-08-19
the ttsc transform envelope stamps no reference graph, so a bundler cache can serve stale generated code


samchon:
## Summary

The `@nestia/core` transform envelope carries only `diagnostics` and `typescript`. It stamps no reference graph, reports no per-file dependencies, and declares no completeness. A bundler that keeps a cache across builds therefore has no way to learn that a generated module depends on a DTO's declaration file, and ttsc's own test suite has a control proving what happens next: the cache replays the stale module after the type changes.

This came out of samchon/ttsc#970, where the same channel was the difference between a correct incremental build and a broken one. typia stamps the graph; nestia does not.

## Evidence

`packages/core/native/transform/transform.go`:

```go
type transformProjectOutput struct {
	Diagnostics []transformCompilerDiagnostic `json:"diagnostics,omitempty"`
	TypeScript  map[string]string             `json:"typescript"`
}
```

A repository-wide search for `NewTransformGraph`, `"graph"`, `dependencies`, `dependenciesComplete`, and `hostInputs` under `packages/core/native` and `packages/sdk/native` returns nothing.

The [ttsc plugin protocol](https://ttsc.dev/docs/development/concepts/protocol) puts the obligation on the producer: the built-in native host and the linked-plugin generic host stamp `graph` automatically, and "an executable sidecar adds the one field to its own envelope". `cmd/ttsc-nestia` and `cmd/ttsc-nestia-sdk` are executable sidecars.

ttsc's own regression suite pins the consequence with a named control:

```
test_webpack_filesystem_cache_control_serves_stale_without_a_graph
  "without a graph-emitting producer, webpack's kept filesystem cache serves
   the stale consumer module after the type file changes"
```

## What it costs a nestia user

1. **Stale generated code in a persistent bundler cache.** A bundler erases type-only imports from its module graph, so nothing tells webpack (NestJS's own `--webpack` builder, and every serverless bundling setup) that `controller.ts`'s generated validator depends on `dto.ts`. Editing the DTO's type leaves the cached module in place. This is a correctness issue, not a performance one.
2. **No watch invalidation through a type-only edge**, for the same reason, in watch mode and in Vite's dev server.
3. **The slowest cache-validation path on every module.** `@ttsc/unplugin` reserves its narrow per-file validation for graph-bearing envelopes. Without a graph it falls back to re-walking and re-comparing the whole project snapshot on every delivered module, which is the most expensive path it has.

## What to change

[...]

PR #1594, opened 2026-07-23
chore(deps): bump typia to ^13.2.0 and ttsc to ^0.20.0

samchon:
## Intent

Bump the `samchon` and `typescript` catalog anchors in `pnpm-workspace.yaml`:

- `typia` (and `@typia/interface`, `@typia/utils` via anchor): `^13.0.2` -> `^13.2.0`
- `ttsc` (and `@ttsc/factory`, `@ttsc/unplugin` via anchor): `^0.18.4` -> `^0.20.0`

`pnpm-lock.yaml` is the corresponding lockfile update.

## Scope

Exactly the two files above; no source changes.

## Verification

No local verification was run; CI validates this pull request. Merge gate: the `build.yml` workflow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1593, opened 2026-07-22
test(sdk): consolidate fixture CI workload

samchon:
## Intent

Implements #1592: remove redundant package-build work from the `sdk` CI job, consolidate compatible compiler-diagnostic fixtures with exact diagnostic oracles, and move duplicated validate-mode coverage to `test-transform-options` while retaining distinct end-to-end SDK boundaries.

## Scope

- Only the `sdk` matrix entry skips the redundant root package build; `tests/test-sdk` retains its explicit five-package self-build.
- Core native diagnostics: 33 fixtures / 33 exact source-scoped diagnostic lines; MCP native diagnostics: 9 fixtures / 11 lines. SDK reflection/report failures remain individual because they occur after compilation.
- The ten body validation-mode fixtures and duplicate header fixture are replaced by the transform-options owner test, including the native header routing table.

## Retained boundaries

The source-tree census retained app/version, clone, Swagger, distribution, watch, CLI, output-layout, native-regression, strictness/nullability, configuration-loading, and WebSocket arity fixtures because they exercise distinct observable boundaries.

## Verification

No local build, test, generator, benchmark, formatter, or dependency installation was run by explicit user direction. Full GitHub Actions verification is running on `28540114`.

## Campaign

Resolves #1592 through the implementation commit's `Close #1592` trailer when this PR merges.

PR #1599, opened 2026-08-19
fix(core): stamp the ttsc transform envelope's reference graph

samchon:
## Intent

Implement issue #1598: the `@nestia/core` transform sidecar publishes an envelope that carries only
`diagnostics` and `typescript`, so a bundler with a persistent filesystem cache has no edge from a
generated validator back to the DTO declaration it was generated from and replays the stale module after
that type changes.

This pull request owns the complete accepted cycle: issue #1598 plus every issue this cycle discovers and
publishes while implementing it. It is opened empty and claims the cycle before any code exists, so the
closing keywords live on the commits that earn them rather than in this body.

## Scope

- [ ] stamp the host-owned reference `graph` in the project-mode transform envelope
- [ ] unify envelope keying on `driver.TransformOutputKey`
- [ ] move the Go-side `ttsc` pin onto the toolchain the npm catalog already resolves
- [ ] `dependencies` / `dependenciesComplete` adjudication with recorded evidence
- [ ] `ttsc` peer-dependency floor
- [ ] regression coverage (Go envelope tests + a ttsc-driven end-to-end envelope assertion)
- [ ] documentation where public behaviour changes

## Verification

Pending. Every step's local evidence and the final matrix land as formal reviews on this thread.


samchon (commented):
### Cycle ledger — `d317f60b`, `b62de3d7`

**`d317f60b` `build(native): pin the Go ttsc modules to the toolchain npm already resolves`**

Moves all fifteen `github.com/samchon/ttsc/...` replacements in
`packages/core/native/go.work`, `packages/sdk/native/go.work`,
`packages/core/test/go.mod` and `packages/sdk/test/go.mod` from
`v0.0.0-20260709053824-84bdc4acccb0` (2026-07-09, pre-0.19.0) to
`v0.0.0-20260723023648-017b4d808689` (the ttsc v0.20.0 release commit), and updates the four
`go.sum` files to match.

Why it is a prerequisite rather than housekeeping: the old module tree has no `driver/graph.go`,
so `driver.NewTransformGraph` does not exist there. A driver-only bump does not compile either —
the 0.20.0 driver reads `shimcompiler.FileAffectsGlobalScope`,
`shimcompiler.GetReferencedFilePaths` and `shimcore.ResolutionMode`, none of which the 2026-07-09
shims export — so the fifteen move together. This was verified by attempting the driver-only bump
first and reading the failure.

The pin also closed a standing skew: `catalog:typescript` already resolves the npm `ttsc` to
0.20.0, which is the toolchain that actually builds this Go source on a consumer's machine, so
local `go build` / `go test` had been compiling against a different driver than production.

`packages/*/native/go.work.sum` is untracked, so CI resolves these hashes through each module's
`go.sum`. Verified locally by removing `go.work.sum` and rebuilding: exit 0, and no `go.work.sum`
was regenerated.

Local evidence: `go build ./...` + `go vet ./...` in `packages/core/native` (exit 0);
`go test ./...` in `packages/core/test` (ok, 19.958s) and `packages/sdk/test` (ok, 44.058s).

**`b62de3d7` `fix(core): stamp the reference graph in the transform envelope`** — resolves #1598 step 1.

- `transformProjectOutput` gains `Graph *driver.TransformGraph \`json:"graph,omitempty"\``.
- `runTransformProject` stamps `driver.NewTransformGraph(prog, cwd)` **before** the per-file loop,
  so the section describes the original sources' resolved references (the transform's inputs)
  rather than the printed result. The loop's `SetParentInChildrenUnset` mutates the node tree it
  walks, which is the concrete reason the order matters; ttsc's own `api-transform` host stamps at
  the same point.
- `sourceFileKey` is deleted in favour of `driver.TransformOutputKey`. Envelope sections are joined
  by key, and the protocol requires one shared implementation; the two were byte-equivalent
  (`filepath.Rel` normalises separators through `Clean` on Windows), so this is a de-duplication,
  not a behaviour change.
- `transform.RunWithOutput` is a new exported seam. Project mode publishes its whole result on
  stdout and, unlike the single-file path, has no `--out`, so an exit code was previously the only
  observable. The alternative — giving the CLI a flag ttsc never passes — would have added product
  surface for a test.

Coverage, with the envelope decoded from the wire shape rather than from the producer's own Go
struct, so a renamed field or a dropped tag fails:

| case | file |
| --- | --- |
| positive: `graph.edges[controller]` carries the type-only DTO edge; `configs` starts at the project tsconfig and reaches its `extends` base | `transform_project_envelope_stamps_reference_graph_test.go` |
| negative twin: a controller that imports no DTO gets no edge to another controller's DTO, and no section names a `bundled:///` library | `transform_project_envelope_graph_omits_an_unreferenced_declaration_test.go` |

Sensitivity proved by mutation: with `output.Graph = ...` commented out, both tests fail
(`project-mode envelope carries no graph section`); restored, both pass.

**Still open in this cycle**: `dependencies` / `dependenciesComplete` adjudication (#1598 steps 2–3),
the `ttsc` peer-dependency floor (step 4), an end-to-end envelope assertion through the real ttsc
host, documentation, and one formatting-drift finding this cycle surfaced (below, once filed).

samchon (commented):
### Cycle ledger — `95981842`, `f8878896`, `679355e8`

**`95981842` `fix(core): declare the ttsc floor and prove the envelope end to end`** — resolves #1598 step 4.

- `@nestia/core` declares `ttsc: ">=0.19.2"` as an **optional** peer. 0.19.2 is where `driver.NewTransformGraph` and `driver.TransformOutputKey` first exist; verified from primary evidence by reading `packages/ttsc/driver/graph.go` at the v0.19.2 release commit `e0efbd386cc2`, which exports both. The two descriptor keys `native/transform.cjs` uses, `composes` and `contributors`, are already in `ITtscPlugin` at that same commit, so the graph SDK is what raises the floor rather than the descriptor shape. Declaring the version this repository happens to pin would have been a compatibility claim narrower than the evidence.
- **Issue premise correction.** #1598 states that "`@nestia/core` and `@nestia/sdk` currently declare no `ttsc` peer dependency at all — ttsc appears only as a devDependency". That holds for `@nestia/core`. It does not hold for `@nestia/sdk`, which carries `ttsc` as a runtime `dependencies` entry at `catalog:typescript` because `TtscExecutor` spawns the binary to compile generated output. A peer entry there would resolve against that own copy and could never warn about the consumer's version, so the floor is declared only where it can be enforced.
- End-to-end coverage in `tests/test-transform-options`, which is the only workspace that drives real `ttsc` over synthetic project files. Every existing case there reads emitted JavaScript, and `graph` never reaches a `.js` file — it reaches the bundler adapter that asked for the transform — so the section was previously unobservable from any test.
- `projectRoot` is load-bearing in that case: without it the project root is `lib/`, every source keys as `../src/...`, the host drops keys that escape cwd, and the assertions would pass over an empty envelope. The case asserts the transformed controller names the DTO before asserting anything about the graph, so an empty envelope cannot read as a pass.
- `setup.mdx` gains the version floor, the bundler-cache consequence, and a troubleshooting line for the plugin-build failure an older ttsc produces.

**`f8878896` `style: restore formatter-clean state and gate it on every pull request`** — resolves #1600, found while running the documented pre-commit `pnpm format`.

Twenty-one tracked files across six packages and six test workspaces were not formatter-clean, and nothing checked. `prettier@3.8.3` has been pinned in `pnpm-lock.yaml` since 2026-07-10 across every lockfile commit since, so this is drift rather than a toolchain artifact.

The gate could not be added first. `pnpm format` also rewrote `packages/migrate/assets/` (cloned by that package's `prepare`) and `packages/migrate/src/bundles/` (emitted from it) — both in `packages/migrate/.gitignore`, neither in `git ls-files` — and the next `pnpm install` reverted the rewrite, leaving `prettier --check` red on nine files a fresh install recreates. `.prettierignore` now excludes both, so the formatter's target set matches the tracked set.

`format.yml` is its own workflow because a `paths:` filter is per workflow and neither `build.yml` nor `test.yml` covers `packages/**/*.ts` + `tests/**/*.ts`; a pull request touching only `tests/**` would skip a job added to either. Proved in both directions locally: the check passes on the reformatted tree, and one deliberate violation fails it from each glob separately — `packages/**` and `tests/**` — because only the second glob is the one that silently regressed before.

**`679355e8` `fix(core): host typia's analysis with the program's own classifications`** — resolves #1601 and #1598 step 2.

Moves the Go `typia` pin from `v0.0.0-20260709134626-fd1c6686950f` (between 13.0.0 and 13.0.1) to `v0.0.0-20260722004126-436fd9275bb5`, the 13.2.0 release commit. `catalog:samchon` has resolved the npm `typia` to `^13.2.0` since #1594, so the generator was two minor versions behind the runtime whose internal helpers its emitted code imports.

*Default-library classification (#1601).* typia grants runtime-native identity only when a runtime authority declares the type, and reads which files those are from a per-checker registry a host installs; with none installed it falls back to a `lib.*.d.ts` base-name test. This host installed nothing, so a project file matching that pattern was taken for a runtime authority and a user-authored `interface Blob { customField: string }` lost its member to `input.blob instanceof Blob` — a check against a constructor the runtime need not provide. Reproduced before the fix and confirmed after: the same fixture emits `instanceof Blob` without the registration and a structural `customField` check with it. Registered in both program-loading hosts (`build`, `transform`) and released before the program closes, so no closed program's checker stays reachable from a process-global map. `check` loads a program but runs no typia analysis, so it is deliberately untouched.

*Consulted declarations (#1598 step 2).* The envelope gains `dependencies`, keyed through `driver.TransformOutputKey` like every other section. The consumer unions it with `graph` (`selectWatchInputs` in `@ttsc/unplugin`), so the list is additive: an omission costs invalidation breadth, never correctness.

*`dependenciesComplete` is deliberately not declared, and this is the one part of #1598 that does not land.* The field transfers responsibility — a listed file stops watching everything the reference closure carries — and this host cannot honour it. Whether a file is transformed at all turns on symbol resolutions no listener reports, including the negative ones: a file whose decorator resolves to something other than `@nestia/core` today is untouched, and an edit to a *different* file can change that resolution without changing this file's text. Declaring such a file complete with an empty list would serve stale output. typia's own producer does not declare it either, and samchon/typia#2357 is that request upstream. The decision is recorded at the fill site rather than left to be rediscovered.

Coverage, both directions, at the two seams:

| seam | positive | negative twin |
| --- | --- | --- |
| `tests/test-transform-options`, real ttsc | a `Blob` declared only by a project `lib.custom.d.ts` under `lib: ["ESNext"]` keeps its member — accepts `{ customField }`, rejects `{}` | the same interface under `lib: ["ESNext", "DOM"]`, where `Blob` is the genuine DOM native, keeps native identity — rejects a structural lookalike, accepts `new Blob([])` |
| `tests/test-transform-options`, envelope | `dependencies[controller]` carries the DTO | a module the transform generated nothing for gets no entry at all |
| `packages/core/test`, wire envelope | `dependencies[TypedBodyController]` carries `IBbsArticle.ts` | `HealthController`, whose route type is a literal, has no entry |

The DOM twin matters on its own: a "fix" that simply stopped classifying natives would satisfy every positive assertion here.

Local evidence at this head: `pnpm test:go` (core 22.878s, sdk cached, both ok), `pnpm --filter ./tests/test-transform-options start` (all seven cases), `pnpm run format:check` clean. The full `pnpm test` and the pull request's own CI are running against this head; results follow.

samchon (commented):
### Cycle ledger — `66b48324`, `b679cbae`, `94c62ab4`, `eb8f8b8a`

**`66b48324` `fix(spell): use the en-us spelling the typos gate enforces`** — CI repair.

`typos.toml` pins `locale = 'en-us'` and `spell-check.yml` gates every pull request; two new comments used "honour" and "behaviour". Swept the whole diff for the same class rather than the two the log named: `git diff master...HEAD | grep -inE "behaviour|honour|colour|favour|normalis|recognis|analys|organis|licence|centre|initialis|serialis|optimis"` now reports only "analysis" and "otherwise".

**`b679cbae` `docs(skills): match the counts to the tree and record the formatting gate`** — resolves #1602.

The counts in the skills are meant to be checked against — the review skill asks for the inventory to be reconciled with `git ls-files` — so a wrong one turns a correct census into a false positive. `transform/` was described as 17 files and has held 13; `test-sdk` was described as 161 feature projects and holds 155. This cycle moves 13 → 15 and 58 → 62 / 77 → 81 tracked `*_test.go`, and every count that already held is left alone: eight packages, seven test workspaces, fifteen `replace` directives of which fourteen are shim redirects, nineteen SDK test files, and five workspaces extending `tests/config/tsconfig.json` directly (the fifth is `test-transform-options` through its `tsconfig.base.json`, which is why the count looked wrong at first glance and is not).

The Commands section now also records `format:check` and `format.yml`.

**`94c62ab4` `fix(migrate): stamp catalog versions into plain dependencies too`** — resolves #1603, found by tracing #1598's ttsc floor into what `@nestia/migrate` generates.

`updateWorkspaceCatalog` resolved a name against both the `samchon` and `typescript` catalogs; `update`, which rewrites every emitted `package.json`, consulted only the first. So the bundler's version ownership depended on which of two equivalent ways a template happened to write a pin. The SDK template writes plain specifiers, so a generated project shipped `ttsc` and `@ttsc/unplugin` at `^0.18.2` beside a `@nestia/core` stamped to this repository's version.

That skew was survivable until this cycle. #1598 raises the floor to `>=0.19.2`, and below it the failure is not a warning: `ttsc` compiles the transform from Go source on the consumer's machine, so the generated project dies in the plugin build. This is the consequence-surface half of the floor change, not a drive-by.

The `??=` in `migratePackageJson` could not have closed it — it fills a missing entry and never aligns a present one, so the comment above it about "the ttsc compiler this repository currently builds against" described an intent the code did not carry out. That line stays: it answers the different question of a template that declares no compiler at all.

Scope of the rewrite, measured before the change: exactly two keys move, `ttsc` and `@ttsc/unplugin`, both `^0.18.2` → `^0.20.0`. `@ttsc/lint` and `@ttsc/paths` stay at the template's `^0.18.2` because this repository pins no version for them, and the nest template is byte-unchanged because it pins through `catalog:typescript`.

`test_migrate_sdk_dependency_catalog_stamp` sits beside the workspace-catalog regression it mirrors. Its oracle is the tracked `pnpm-workspace.yaml` declaration, not the lockfile the bundler itself reads, so it compares the output against what the repository declares rather than against the bundler's own input. Its negative twin asserts `tinyglobby` — pinned nowhere here — keeps the template's version, so a rewrite that overwrote every specifier would not pass. Proved sensitive by mutation: with the emitted bundle put back to `^0.18.2` the case fails with the expected message and passes again once restored.

**`eb8f8b8a` `test(core): keep the envelope helpers in the helper file`** — review cleanup, no behaviour change.

`mustOmitBundledScheme` and `keysOf` were defined inside the case files that first needed them, against this module's one-case-per-file convention; the second case to need either would have had to import from a case file. Also corrects a doc comment that still named `sourceFileKey`, the local helper this cycle replaced with `driver.TransformOutputKey`.

### Cycle scope, closed

| issue | disposition |
| --- | --- |
| #1598 | steps 1, 2 and 4 implemented; step 3 recorded as not declarable by this host, with the reason at the fill site |
| #1600 | fixed |
| #1601 | fixed |
| #1602 | fixed |
| #1603 | fixed |

No further discovery rounds are being run: the mission scoped new issues to what implementing #1598 surfaced, and #1600–#1603 are exactly that set.

Local evidence at `eb8f8b8a`: `go test ./...` in `packages/core/test` (ok), `pnpm --filter ./tests/test-migrate start` (all features), `pnpm run format:check` clean. The full `pnpm test` and this pull request's CI are running against this head.

samchon (commented):
# HANDOFF — read this first to resume this pull request

Everything below is state, not narrative. The branch is `fix/transform-graph-envelope`, head `eb8f8b8a`, worktree clean, all work pushed. Nothing is half-written on anyone's disk.

## 1. What is done

| issue | status | where |
| --- | --- | --- |
| #1598 | steps 1, 2, 4 implemented; step 3 deliberately not implemented (see §4) | `b62de3d7`, `95981842`, `679355e8` |
| #1600 | implemented | `f8878896` |
| #1601 | implemented | `679355e8` |
| #1602 | implemented | `b679cbae` |
| #1603 | implemented | `94c62ab4` |

All five issues are closed as development-complete. Each commit carries its own `Close #n` line, so a squash merge closes them again idempotently; `git log origin/master..HEAD` was reconciled and contains no revert, so every closing line has a surviving fix at `HEAD`.

Commits, oldest first:

```
0bc4cad6 chore(campaign): claim the transform-envelope reference-graph cycle   (empty claim)
d317f60b build(native): pin the Go ttsc modules to the toolchain npm already resolves
b62de3d7 fix(core): stamp the reference graph in the transform envelope        Close #1598
95981842 fix(core): declare the ttsc floor and prove the envelope end to end
f8878896 style: restore formatter-clean state and gate it on every pull request Close #1600
679355e8 fix(core): host typia's analysis with the program's own classifications Close #1601
66b48324 fix(spell): use the en-us spelling the typos gate enforces
b679cbae docs(skills): match the counts to the tree and record the formatting gate Close #1602
94c62ab4 fix(migrate): stamp catalog versions into plain dependencies too      Close #1603
eb8f8b8a test(core): keep the envelope helpers in the helper file
```

## 2. The change, file by file

**Envelope (the point of #1598).** `packages/core/native/transform/transform.go`

- `transformProjectOutput` gained `Graph *driver.TransformGraph` (`json:"graph,omitempty"`) and `Dependencies map[string][]string` (`json:"dependencies,omitempty"`).
- `runTransformProject` stamps `driver.NewTransformGraph(prog, cwd)` **before** the per-file loop. Order is load-bearing: the graph must describe the original sources' resolved references, and the loop's `SetParentInChildrenUnset` mutates the node tree it walks. ttsc's own `api-transform` host stamps at the same point.
- `sourceFileKey` was deleted; keys now come from `driver.TransformOutputKey`, the one implementation the protocol requires every section to share. The two were byte-equivalent (`filepath.Rel` normalises separators through `Clean` on Windows), so this is de-duplication, not a behaviour change.
- The dependency collector wraps the loop with `Begin(key)` / `End()` and `schemametadata.MetadataDependency_listen(prog.Checker, collector.Touch)`.

`packages/core/native/transform/typia_dependency.go` (new) — the collector. Ported from typia's own producer shape: memoised per-file-name values, drops `bundled:///`-style virtual sources, drops default libraries **by asking the program** (`prog.TSProgram.IsLibFile`) rather than by file name, drops the transformed file itself, sorts values, returns `nil` when empty so the field stays off the wire.

`packages/core/native/transform/typia_registry.go` (new) — `registerTypiaDefaultLibraryClassifier(prog) func()`, the fix for #1601. Registered in **both** program-loading hosts (`runBuild` in `build.go`, `runTransform` in `transform.go`) and released before `prog.Close()` runs, because the registry is keyed by checker and a process-global map must not retain a closed program's checker. `runCheck` loads a program but runs no typia analysis, so it is deliberately untouched.

`packages/core/native/transform/run.go` — `RunWithOutput(args, out, errOut)`, a test seam. Project mode publishes its whole result on stdout and, unlike the single-file path, has no `--out`, so an exit code was previously the only observable. The alternative was giving the CLI a flag ttsc never passes.

**Manifest and docs.** `packages/core/package.json` declares `ttsc: ">=0.19.2"` as an **optional** peer. `@nestia/sdk` deliberately does not get one: it already carries `ttsc` as a runtime dependency, so a peer there would resolve against its own copy and could never warn about the consumer's. `website/src/content/docs/setup.mdx` gained the floor, the bundler-cache consequence, and a troubleshooting line.

**Formatting (#1600).** `.prettierignore` excludes `packages/migrate/assets/` and `packages/migrate/src/bundles/` (both regenerated by that package's `prepare`, both in `packages/migrate/.gitignore`, neither in `git ls-files`); 21 drifted tracked files reformatted; `format:check` added to the root scripts; `.github/workflows/format.yml` runs it. It is its own workflow because a `paths:` filter is per workflow and neither `build.yml` nor `test.yml` covers the whole `packages/**/*.ts` + `tests/**/*.ts` surface.

**Migrate bundler (#1603).** `packages/migrate/src/executable/bundle.js::update` now resolves a plain specifier against both catalogs in the same order `updateWorkspaceCatalog` already did. Exactly two keys move in the emitted SDK template: `ttsc` and `@ttsc/unplugin`, `^0.18.2` → `^0.20.0`. `@ttsc/lint` / `@ttsc/paths` stay put (this repository pins no version for them) and the nest template is byte-unchanged (it pins through `catalog:typescript`).

**Skills (#1602).** `transform/` 17 → 15 (it had been 13 before this cycle added two files), `test-sdk` 161 → 155, `packages/core/test` 58 → 62, tracked `*_test.go` 77 → 81. Counts that already held were left alone. `format:check` / `format.yml` recorded in the Commands section.

## 3. Dependency pins — the part most likely to confuse a reader

Two Go pins moved, in `packages/core/native/go.work`, `packages/sdk/native/go.work`, `packages/core/test/go.mod`, `packages/sdk/test/go.mod`, plus all four `go.sum`.

**ttsc**: `v0.0.0-20260709053824-84bdc4acccb0` (2026-07-09, pre-0.19.0) → `v0.0.0-20260723023648-017b4d808689` (the v0.20.0 release commit). Required: the old module tree has no `driver/graph.go` at all. A driver-only bump does **not** compile — the 0.20.0 driver reads `shimcompiler.FileAffectsGlobalScope`, `shimcompiler.GetReferencedFilePaths` and `shimcore.ResolutionMode`, none exported by the 2026-07-09 shims. This was verified by attempting the driver-only bump and reading the failure, so do not try to shrink the change to one line. It also closed a standing skew: `catalog:typescript` already resolved the npm `ttsc` to 0.20.0, which is the toolchain that actually compiles this Go source on a consumer's machine.

**typia**: `v0.0.0-20260709134626-fd1c6686950f` (between 13.0.0 and 13.0.1) → `v0.0.0-20260722004126-436fd9275bb5` (13.2.0). Required for `MetadataDependency_listen` (#1598 step 2) and it is what makes `MetadataDefaultLibrary_register` available at all (#1601). `catalog:samchon` has resolved the npm `typia` to `^13.2.0` since #1594, so this closed a generator/runtime skew rather than opening one.

**go.sum trap.** `packages/*/native/go.work.sum` is untracked, so CI resolves these hashes through each module's `go.sum`. All four were updated. Verified by deleting `go.work.sum` and rebuilding: exit 0, and it was not regenerated. If you move a pin again, do the same check — `go mod tidy` cannot help here, it disables workspace mode and the shim replacements vanish.

**Go proxy lag.** proxy.golang.org 404s for a freshly requested pseudo-version of the ttsc shim submodules and backfills minutes later. If a bump 404s, request it once and retry rather than concluding the version does not exist.

## 4. `dependenciesComplete` — why it is not declared, and what would unblock it

This is the one part of #1598 that does not land, and it is a decision rather than an omission. The reason is recorded at the fill site in `runTransformProject` so it is not rediscovered as a gap.

The field narrows a consumer's per-file bound from `dependencies[F] ∪ reach(edges,F) ∪ globals ∪ configs` to `dependencies[F] ∪ configs`, so an omission serves **stale output** — it is a responsibility transfer, not a hint. This host cannot make the claim:

1. Whether a file is transformed at all turns on symbol resolutions no listener reports, including the negative ones. A file whose decorator resolves to something other than `@nestia/core` today is untouched; editing a *different* file so that identifier re-exports `TypedBody` changes that with this file's own text unchanged. Declaring it complete with an empty list would serve the untransformed output forever.
2. typia's trigger set belongs to typia. This host runs typia's transformer; the set of call shapes typia rewrites can change with a typia release, and a completeness claim made here would silently become wrong.

Unblocking needs both of: samchon/typia#2357 (OPEN — the same request on typia's side), and the untouched-file rule from samchon/ttsc#1259 (CLOSED 2026-08-19, but in the 0.28 line, while this repository pins `^0.20.0`). When those land, the collector in `typia_dependency.go` is already the right substrate — typia 14's own collector adds an `Unbound()` / `unbounded` set for exactly this, and porting that shape is the next step.

## 5. Verification state at `eb8f8b8a`

**Green already:** `Ubuntu` (build), `format`, `spell-check`, `deploy` (the full website build including TypeDoc), `benchmark-system (windows-latest)`, CodeQL ×3, Socket ×2.

**Still pending when this was written:** `go`, `sdk`, `migrate`, `transform-options`, `e2e`, `cli`, `editor`, `benchmark`, `benchmark-system (ubuntu-latest)`. Read them at https://github.com/samchon/nestia/actions/runs/32272710830 before doing anything else.

**Run locally, all passing, at or just before this head:**

- `pnpm test:go` — `packages/core/test` ok, `packages/sdk/test` ok
- `pnpm --filter ./tests/test-transform-options start` — all seven cases including the two new ones
- `pnpm --filter ./tests/test-migrate start` — all features including the new one
- `pnpm run format:check` — clean
- `go build ./...` and `go vet ./...` in `packages/core/native`

A full `pnpm test` was running against this exact head and had not reached the test workspaces when work stopped. **That is the one gap.** It is not a known failure — it is an unfinished run. CI's matrix covers the same ground per workspace, so the honest way to close the gap is either to let those nine lanes settle or to run `pnpm test` once locally.

**Mutation records — every product change was proved sensitive:**

| change | mutation | result |
| --- | --- | --- |
| `output.Graph = driver.NewTransformGraph(...)` | commented out | both Go envelope cases fail with "project-mode envelope carries no graph section"; pass on restore |
| default-library classifier registration | replaced with a no-op release | the same fixture emits `input.blob instanceof Blob` and drops the `customField` check; restored, it emits the structural check |
| `bundle.js` catalog stamping | emitted bundle put back to `ttsc ^0.18.2` | `test_migrate_sdk_dependency_catalog_stamp` fails with "must pin ttsc at ^0.20.0, got ^0.18.2"; pass on restore |
| `.prettierignore` + `format:check` | one deliberate violation per glob | the check fails from `packages/**` and from `tests/**` separately; clean on restore |

## 6. Tests added, and what each one is for

| file | asserts |
| --- | --- |
| `packages/core/test/transform_project_envelope_stamps_reference_graph_test.go` | the envelope carries `graph`; `edges[TypedBodyController.ts]` holds `src/api/structures/IBbsArticle.ts` (a type-only edge); `configs` starts at the project tsconfig and reaches the extended base |
| `packages/core/test/transform_project_envelope_graph_omits_an_unreferenced_declaration_test.go` | negative twin — `HealthController`, which imports no DTO, has no edge to that DTO, and no section names a `bundled:///` path |
| `packages/core/test/transform_project_envelope_reports_consulted_declarations_test.go` | `dependencies[TypedBodyController.ts]` holds the DTO; `HealthController` gets no entry at all |
| `packages/core/test/helpers_project_envelope_test.go` | the shared harness. It decodes the envelope from the **wire shape**, not from the producer's Go struct, so a renamed field or a dropped tag fails |
| `tests/test-transform-options` `transform envelope graph` | the same properties end to end through real `ttsc` via `TtscCompiler.transform()`, plus the `dependencies` positive and its negative twin |
| `tests/test-transform-options` `user-authored global keeps its members` | #1601 both directions: a `Blob` declared only by a project `lib.custom.d.ts` under `lib: ["ESNext"]` keeps `customField`; the one-axis twin under `lib: ["ESNext","DOM"]`, where `Blob` is the real DOM native, keeps `instanceof` identity |
| `tests/test-migrate/src/features/test_migrate_sdk_dependency_catalog_stamp.ts` | #1603, mirroring the existing workspace-catalog case; oracle is the tracked `pnpm-workspace.yaml`, negative twin is `tinyglobby` staying untouched |

Two traps a future editor of these tests will hit:

- In the `test-transform-options` envelope case, `projectRoot: __dirname` is load-bearing. Without it the project root is `lib/`, every source keys as `../src/...`, the native host drops keys that escape cwd, and the envelope comes back **empty** — every assertion would then pass vacuously. That is why the case asserts the transformed controller names the DTO *before* asserting anything about the graph.
- The `native-global-user` project must keep `lib: ["ESNext"]` **and** `types: []`. Add `@types/node` back and Node's own `Blob` becomes a genuine runtime authority, the divergence disappears, and the case silently stops testing anything.

## 7. What remains before merge

1. Read the nine pending lanes. Fix anything red in this same pull request, including a failure that predates the cycle.
2. Run one complete Self-Review round over the whole `master...HEAD` diff at the settled head. Three ledger reviews are already on this thread; they are ledgers, not that round.
3. Integration-sensitive gate — the native Go transform, package manifests, and CLI code all changed, so construct the prospective merge of `origin/master` and this head in a disposable checkout and run root build, `pnpm test`, packaging and clean packed-consumer checks there; re-read `origin/master` first and restart the gate if it moved.
4. Merge with the repository's established method. Do **not** touch any `version` field — this cycle freezes versions, and the native change earns its version from a separate maintainer release commit.
5. After merge: verify the five issues are closed against surviving fixes, delete the topic branch, and remove the disposable roots this cycle created (`tests/test-migrate/.generated`, the regenerated `tests/test-sdk` output, `deploy/tarballs/`, any mutation checkout).

## 8. Things checked that turned out to be non-issues

Recorded so nobody re-derives them: the `benchmark` workspace's `test` script asserts only worker-selection logic and never compares committed `benchmark/results/**`, so the typia bump cannot move it. `deploy/verify-package-exports.cjs` and `release-guard.cjs` are unaffected by a peer-dependency addition. `pnpm install --frozen-lockfile` still succeeds after the peer was added, because pnpm's lockfile `importers` do not record workspace peers. The `README.md` mentions `ttsc` nowhere, so the floor belongs only in `setup.mdx`. `packages/sdk/test` drives programs without the classifier registered, which is a harness/production divergence with no behavioural difference for those fixtures — it is noted, not fixed.

samchon (commented):
### Correction — this supersedes the handoff review above

The handoff comment was written at `eb8f8b8a` and is now wrong in its first table. Three commits landed after it and one of them is a revert. Read this instead.

### `897af4ab` `Revert "fix(core): host typia's analysis with the program's own classifications"`

The Go `typia` pin move from 13.0.x to 13.2.0 is out, and with it the two things that needed it. CI is what forced the decision: the `sdk` lane failed on four features — `clone-and-keyword`, `clone-and-propagate`, `clone-and-propagate-and-keyword`, `clone-implicit` — and none of them is about the transform.

typia 13.2.0 deliberately changed the separator it mints for a duplicated metadata name from `.` to `-o`, and says why in `MetadataCollection.go`:

> It must not be ".", which is how a qualified name separates a namespace from its member. Two consumers read that meaning back out of an allocated id ... A dotted counter also let an invented id land on the real full name of a `namespace Foo { interface o1 }` member, and the loser was dropped from the document outright.

`@nestia/sdk` is one of those consumers. `SdkHttpCloneProgrammer.prepare` splits a metadata name on `.` and treats every segment as a TypeScript identifier and a file name, so `IDirectory.o1` used to render as `namespace IDirectory { type o1 }` and `IDirectory-o1` now renders as `import type { IDirectory-o1 }`, which does not parse. `StringUtil.isImplicit` has the same assumption as literal prefixes (`__type.`, `__object.`), so an anonymous type that used to be filtered out now asks to be declared.

Filed as #1604 rather than fixed here: the fix means choosing what a generated SDK calls a disambiguated type, which renames published generated output and is the maintainer's call, not a side effect of a cache-invalidation issue.

What the revert costs, and nothing else:

- **#1598 step 2 (`dependencies`)** needs `MetadataDependency_listen`, which is 13.2.0-only. It is additive by contract — the consumer unions it with `graph` — and for this producer the reported set is a subset of the graph's reachability closure, so its absence changes no invalidation decision. Only a `dependenciesComplete` declaration would, and this envelope was never going to make one (#1598 comment explains why).
- **#1601** needs `MetadataDefaultLibrary_register`, which does not exist in the pinned module at all, so the defect is neither reachable nor fixable there. Reopened and blocked on #1604. The implementation and its two-direction fixture survive in the reverted commit and come back with `git revert 897af4ab`.

### `42f9c5b6` + `5a898e62` — the `sdk` lane's pre-existing failure

`openapi_v2` has been red on master through at least #1594 and #1597. It was two defects in a chain, and neither alone makes the feature pass.

**#1605** — nestia's *own* default server carries `description: "insert your server url"`, Swagger 2.0 has nowhere to put a server description, and `@typia/utils`'s downgrader refuses to discard one. So `swagger.openapi: "2.0"`, a documented `INestiaConfig` option, was unreachable for any project that did not override `swagger.servers`: generation aborted on data the user never wrote. Refusing is the right contract for a description the *user* wrote, so only nestia's own default changes, and only for the one target version that cannot carry it.

**#1606** — a `void` route got `content: { "application/json": {} }`: a media type with no schema, which says "this endpoint replies with JSON" and then declines to say what the JSON is. A client generator reads that as a body and emits a parse for one that never arrives. `content` now exists exactly when there is a body to describe, tested on what the success metadata yields rather than on how the return type is spelled, so `void`, `undefined` and `never` are one case. An example without a schema keeps the entry; a `binary` route keeps its `{ type: "string", format: "binary" }`.

Verified: `pnpm --filter ./tests/test-sdk start -- --only openapi` passes both `openapi_v2` and `openapi_v3`. Checked that no existing swagger assertion reads a `void` route's `content` — `status`, `exception`, `clone-and-propagate` and `swagger-example` all assert on responses that carry a real schema.

### Self-review finding: a closing keyword outlives its fix

`git log origin/master..HEAD` shows `679355e8` still carrying `Close #1601`, and `897af4ab` reverted it. A squash merge concatenates every commit message, so **the merge will close #1601 with no surviving fix at `HEAD`**. That is mechanical, not a decision.

The merge gate must reopen #1601 immediately after merge and say the merge closed it mechanically. Every other closing line has a surviving fix, reconciled against the same range: #1598 (graph), #1600, #1602, #1603, #1605, #1606.

### Cycle scope, final

| issue | in this pull request |
| --- | --- |
| #1598 | step 1 (`graph`) and step 4 (`ttsc` peer floor). Step 2 blocked on #1604; step 3 not declarable by this host, reason recorded at the fill site |
| #1600 | fixed |
| #1601 | **reverted**, reopened, blocked on #1604 |
| #1602 | fixed |
| #1603 | fixed |
| #1604 | filed with full diagnosis, not fixed — renames generated output |
| #1605 | fixed |
| #1606 | fixed |

Local evidence at `5a898e62`: `pnpm --filter ./packages/sdk build`, `pnpm --filter ./tests/test-sdk start -- --only openapi`, `pnpm run format:check`, `go test ./...` in `packages/core/test`. Full `test-sdk` and this pull request's CI are running against this head.

samchon (commented):
# Self-Review round 1 — complete pass over `origin/master...HEAD`

One reviewer, whole surface, fresh start from `a11f7167...dbeb1b3d`. Every changed file and hunk was read, including the ones the three ledger comments above already describe; a ledger is not a round. Five findings survived verification and are applied in `cd02e681`, `022d14bb`, `b9d35c5d`, `b0ea0955` and `7dfddc75`.

## What the round covered

51 changed files, grouped by what needed proving rather than by directory:

| surface | how it was checked |
| --- | --- |
| Go transform envelope (`transform.go`, `run.go`) | read; `driver.TransformOutputKey` read at its source and compared against the deleted `sourceFileKey` |
| Go module pins | `go.work` / `go.mod` / `go.sum` reconciled across all four modules; typia confirmed fully restored to the pre-revert pin |
| `ttsc >=0.19.2` peer floor | rebuilt against a real 0.19.2 tarball (below) |
| `packages/sdk` swagger composition | read; the downgrader probed directly for every media-type shape |
| `packages/migrate` bundler | read against `updateWorkspaceCatalog`, whose resolution order it now matches |
| formatting, workflow, manifest | read; `format:check` and `gofmt -l` run over the whole tree |
| skills and website prose | reconciled against `git ls-files` and against `INestiaConfig` |
| the revert | `679355e8`'s whole file list confirmed absent at `HEAD` |

## Findings

### 1. The revert left three counts describing a tree that no longer exists — `cd02e681`

`b679cbae` stamped the counts the cycle was going to produce, and `897af4ab` then reverted the commit that produced three of the files those counts covered. At `HEAD` the `transform/` tree is 13 files and was documented as 15, `packages/core/test` holds 61 `*_test.go` and was documented as 62, and the repository holds 80 tracked `*_test.go` and was documented as 81.

Not cosmetic in the way a stale number usually is: the review skill asks for the inventory to be reconciled with `git ls-files`, so these are read as an oracle and a wrong one reports a correct census as a discrepancy. Every other count in both skills was re-derived and holds, so none of them moves.

### 2. One Go file in the tree is not gofmt-clean, and this pull request added it — `022d14bb`

`gofmt -l` over all 100 tracked `*.go` files — read from the blobs, so a CRLF working copy does not confound it — reports exactly one: `helpers_project_envelope_test.go`, three struct tags one column past gofmt's alignment. Nothing in CI runs `gofmt`, so it would have stayed.

### 3. The swagger guide documents the wrong value set and default for `swagger.openapi` — `b9d35c5d`, filed as #1608

The page states the option twice and both statements were wrong the same two ways. `INestiaConfig.openapi` is `"2.0" | "3.0" | "3.1" | "3.2"`, and both `SwaggerGenerator` and `NestiaSwaggerComposer` read `config.openapi ?? "3.2"`, so the value most projects actually get was documented nowhere and the page named a default the code does not produce. Pre-existing, but it is the page documenting the option #1605 made reachable, and a reader arriving because of that fix lands on it.

### 4. #1606's fix stopped one loop short — `b0ea0955`, filed as #1607

This is the substantive finding. `5a898e62` removed the schemaless media type from a route's success response. The same function writes one response object per declared exception, and that loop still decided `content` on nothing at all. Reproduced end to end by adding one decorator to the fixture #1606 was verified against, at `dbeb1b3d` with both #1605 and #1606 already in:

```ts
@core.TypedException<void>({ status: 400 })
@core.TypedRoute.Get()
public get(): void {}
```

```
Generating Swagger Document
SwaggerV2Downgrader: response media types require a shared schema or example.
  - openapi_v2: failed
```

Same mechanism, one loop earlier in the same function: `props.schema` is a lookup into metadata that `SwaggerGenerator.compose` filters with `sizeOf(m) !== 0`, so a `void` exception type never enters the dictionary and the lookup returns `undefined`. `swagger.openapi: "2.0"` was therefore still unreachable for any project documenting a bodyless error — the failure both earlier fixes were for.

The fix moves the rule into one helper both response kinds go through rather than writing it a second time. Probing the downgrader directly turned the boundaries into branches rather than guesses:

| response media type | `downgradeDocument(doc, "2.0")` |
| --- | --- |
| `{}` on `200`, and `{}` on `400` | throws `response media types require a shared schema or example` |
| `{ schema: {} }` — the `@throws` JSDoc path | accepted |
| `{ example: 1 }` | accepted |
| `{ examples: {...} }`, with or without a schema | throws `response named examples are not representable` |
| no `content` key | accepted |

The fourth row is worth recording as a non-issue: named examples are refused at 2.0 whether or not a schema is present, so that refusal is the downgrader's own contract about examples and is untouched by this rule.

### 5. #1605 and #1606 landed with no assertion on what they produce — `b0ea0955` and `7dfddc75`

`openapi_v2` asserts only that the emitted file validates as a Swagger 2.0 document, so a passing feature proved generation stopped throwing. A document that never reached the downgrader would satisfy that too, and nothing pinned the response object, the absence of `content` at 3.x, or what the default server becomes at either version. #1606's own acceptance criteria asked for exactly this and `5a898e62` did not add it.

Four cases now, each with its twin one property away: the `void` route and its `void` exception declare no body while a typed exception beside them and a payload route still do; at 3.0 the same absence holds where nothing refuses the document, so the rule is not narrowed to the version that complains; and the default server splits into `host` / `basePath` / `schemes` at 2.0 while keeping its description at 3.0, so a fix that simply stopped emitting the description would fail.

`openapi_v2`'s `HealthController` is changed rather than a feature added: it is the fixture both issues were verified against and the only project targeting 2.0, so one document now pins both directions through the downgrade.

## `ttsc >=0.19.2` — verified from primary evidence, both directions

The floor is a published compatibility claim in `peerDependencies` and in `setup.mdx`, and symbol existence alone does not establish it. `packages/*/native/go.work` is not in either package's `files`, and ttsc refuses a plugin `go.mod` that replaces a ttsc-managed module, generating its own workspace from the installed npm ttsc's Go tree. So the consumer's ttsc version, not this repository's pins, is what the plugin compiles against, and the real claim is that nestia's Go source builds against 0.19.2's driver and shims.

Rebuilt to check, replicating what `buildSourcePlugin` generates, against the real `ttsc@0.19.2` tarball:

- `packages/core/native` — `go build ./...` exit 0
- `packages/sdk/native` — `go build ./...` exit 0
- `ttsc@0.19.2` ships `driver/graph.go` exporting both `NewTransformGraph` and `TransformOutputKey`, and its host parses the `graph` envelope section (`lib/compiler/internal/transformProjectInMemory.js`)
- `ttsc@0.19.1` and `0.19.0` have no `driver/graph.go` at all, so the floor is exact rather than conservative, and `setup.mdx`'s "fails the plugin build on an unknown Go symbol" is the real failure mode

Also confirmed while there: `driver.TransformOutputKey` is byte-identical in 0.19.2 and 0.20.0 and behaviourally identical to the deleted `sourceFileKey`, so `b62de3d7`'s de-duplication claim holds.

## Checked, not findings

- The typia revert is complete. `679355e8`'s whole file list — `typia_dependency.go`, `typia_registry.go`, `transform_project_envelope_reports_consulted_declarations_test.go`, the `native-global` fixtures — is absent at `HEAD`, all four `go.sum` files carry only the old typia hash, and every residual diff across that range belongs to an intervening commit.
- `pnpm install` in a clean checkout of `dbeb1b3d` leaves `pnpm-lock.yaml` untouched, confirming the peer addition needs no lockfile change.
- The `graph.configs[0] === "lib/envelope.json"` assertion is separator-safe: `TransformOutputKey` ends in `filepath.ToSlash`.
- `props.schema` is a `WeakMap` lookup with no side effects, so hoisting the call out of the ternary changes no ordering.
- `@nestia/migrate` consumes the success response only, so the exception-response change has no consumer there.
- The two `.prettierignore` additions are absent from `git ls-files`, as the entry claims.
- `format.yml` carrying no `concurrency` group matches `build.yml`, `spell-check.yml` and `website.yml`; only `test.yml` has one.
- `setup.mdx`'s new paragraphs are one source line each, which is the documentation skill's rule; the hard-wrapped text around them is older.
- All fifteen `replace` directives, the fourteen shim redirects, the eight packages, the seven test workspaces, the 155 `test-sdk` feature projects, the 19 SDK Go test files, and the five workspaces extending `tests/config/tsconfig.json` directly were re-counted and every one of those documented numbers holds.

## Recorded for a later round, outside this diff

`SwaggerOperationComposer` writes `requestBody` with `props.schema(props.route.body.metadata)!`. That is the same lookup that returns `undefined` for size-0 metadata, so a bodyless request body would produce the same schemaless media type. This pull request does not touch that path and it is no more reachable than before, and whether `@TypedBody() input: void` is diagnosed upstream was not resolved here. Recorded rather than claimed safe.

## Verification at this round

- finding 4 reproduced before the fix, and `pnpm --filter ./tests/test-sdk start -- --only openapi` green after it, twice: once with the `void` exception alone and again with the typed twin beside it
- `node ../../node_modules/typescript/bin/tsc --noEmit -p tsconfig.json` in `packages/sdk`
- `pnpm run format` then `pnpm run format:check` — clean
- `gofmt -l` over every tracked `*.go` blob — clean after the fix
- `go build ./...` for both native modules against `ttsc@0.19.2`

The integration-sensitive gate and a second complete round run against the settled head; results follow on this thread.

samchon (commented):
# Self-Review round 2 — fresh pass over `origin/master...7dfddc75`

Round 1 applied five findings, so this is the full round the law requires afterwards, not a recheck of the fixes. Whole surface again, 57 changed files, every one read — including the five commits round 1 produced, which no round had reviewed.

One finding.

## Finding: the helper's invariant was a claim, not a structure — `68600ba7`

`b0ea0955` moved the "a response carries `content` exactly when it has a body" rule into one helper and documented it as the single place a response media type is written. Two of the three sites went through it. The `@throws` JSDoc path still built its own:

```ts
content: {
  "application/json": {
    schema: {},
  },
},
```

Nothing was wrong with the output — `{}` is a schema, so the rule keeps that entry either way, and the serialized document is identical. What was wrong is that the invariant held by coincidence at that site. The next edit to what counts as a described body would have moved two sites and missed the third, which is precisely the shape of the defect this cycle just fixed: `5a898e62` changed the success response and left the exception loop behind, and the loop kept producing the document the fix existed to remove.

`68600ba7` routes it through the helper and records at the call site why `{}` counts as a description: the tag names a status and nothing about its payload, so the response says a body exists without constraining it.

Verified on Node 24: `pnpm --filter ./tests/test-sdk start -- --only exception` — `exception`, `exception-filter`, `exception-error-bigint`, `exception-with-propagate` all green. The `exception` feature is the only fixture that drives `@throws` tags, and it asserts on the emitted responses.

## Re-verified at this head, not carried over from round 1

- All 57 changed files read again, and reconciled against `git diff --name-status`; nothing in the diff went unexamined.
- No `version` field appears anywhere in the diff, as [Change Integrity](https://github.com/samchon/nestia/blob/master/.agents/skills/development/SKILL.md#change-integrity) requires of an implementation pull request.
- No `.env`, `deploy/tarballs/`, `.generated`, `BENCHMARK.md`, generated `src/api/functional`, or `swagger.json` is tracked by the diff; `benchmark/results/**` is untouched.
- `setup.mdx`'s claim that the bundler adapter registers the DTO as an input was checked against the consumer rather than assumed: `@ttsc/unplugin@0.20.0`'s `selectWatchInputs` reads `props.result.graph` and feeds `selectGraphInputs` and `selectResolutionCandidateInputs` from it, which is exactly the `edges` / `candidates` the envelope now stamps.
- The skill counts corrected in `cd02e681` were re-derived from `git ls-files` a second time at this head.

## The one thing round 1 recorded and this round settled

Round 1 recorded that `SwaggerOperationComposer` writes `requestBody` through `props.schema(...)!`, the same lookup that returns `undefined` for size-0 metadata. It stays recorded rather than fixed: `SwaggerOperationParameterComposer.body` is the only remaining site that builds a media type by hand, it is reached only when `route.body !== null`, this pull request does not touch it, and whether `@TypedBody() input: void` survives the transform's diagnostics was not established. A fix without that answer would be a guess about reachability.

samchon (commented):
# Self-Review round 3, the merge gate, and the closing-keyword reconciliation

Round 2 produced `68600ba7`, so this is the fresh complete round after it. Whole surface again, 57 changed files, read at this head. **No finding survived verification in the diff itself.** The merge gate below then found one the diff could not show — #1611, a Windows portability defect in a fixture this pull request does not touch — so a fourth round follows the fix rather than this being the last one.

## Round 3

Re-derived rather than carried forward:

- every changed file read again against `git diff --name-status origin/master...HEAD`, all 57 accounted for
- `git ls-files` re-counted for all three numbers `cd02e681` corrected, plus the fifteen `replace` directives, fourteen shim redirects, eight packages, seven test workspaces, 155 `test-sdk` feature projects, 19 SDK Go test files, and five workspaces extending `tests/config/tsconfig.json` — every documented count holds
- no `version` field anywhere in the diff
- nothing tracked under `.env`, `deploy/tarballs/`, `.generated`, `BENCHMARK.md`, `src/api/functional`, `swagger.json`, or `benchmark/results/**`
- `pnpm run format:check` clean, and `gofmt -l` over all 100 tracked `*.go` blobs clean
- one observation from running the first of those on Windows became #1610, below
- the composer read end to end at its final shape: three response kinds, one helper, no site left writing a media type by hand

## One nuance the round surfaced and deliberately did not change

`driver.TransformGraph` gained `Candidates` in ttsc 0.20.0; at 0.19.2, the declared floor, the struct carries only `Edges`, `Globals` and `Configs`. The `Graph` doc comment in `transform.go` lists the resolution candidates among what the section carries, which is exact at the pinned 0.20.0 and one section short at the floor.

It is left alone, and the floor is unaffected either way: nestia never names `Candidates` — it assigns whatever `driver.NewTransformGraph` returns — which is why both native modules build against 0.19.2, and `@ttsc/unplugin` already returns an empty list when `graph.candidates` is absent. `setup.mdx`'s correctness claim rests on `edges`, which 0.19.2 has. Naming a dependency's version inside a comment about that dependency's struct would age worse than the sentence does, and the alternative — a fresh full review round for a clause with no behavioural consequence — buys nothing.

Related and confirmed while reading that file: stamping the graph *before* the transform loop is load-bearing, not stylistic. `NewTransformGraph` reads the AST through `shimcompiler.FileAffectsGlobalScope` and `referenceTargets`, and the loop's `SetParentInChildrenUnset` mutates the node tree those walk.

## The helper's two example clauses, traced for reachability rather than assumed

`described` accepts on a schema, on `example`, or on a non-empty `examples`. The last two only matter when the schema is absent, so this round asked whether either is reachable through the public decorator surface rather than adding a case on the assumption that they are.

`TypedException.IProps<T>` types `example?: T | undefined`, so for the bodyless types that produce no schema — `void`, `undefined`, `never` — the only value that type-checks *is* `undefined`, and the `example` clause cannot decide anything. It stays as the defensive half of a rule stated once for all three response kinds, not as a branch with a missing test.

`examples?: Record<string, IExample<T>>` is different: `IExample<void>.value` accepts `undefined`, so a bodyless exception can carry named examples, `described` accepts, and the entry is written with a schemaless media type and real `examples`. At 3.x that document is legal and says what it means. At 2.0 it is refused — but by `response named examples are not representable`, which the round-1 probe established is the downgrader's contract about named examples generally, refused with or without a schema. So it is not the schemaless-media-type defect reappearing, and narrowing `described` to exclude it would delete information the user wrote.

## Mutation records

Every product change in this cycle is now proved sensitive. The four from the earlier ledgers stand; these are the two this review round added, both run on Node 24 with the tests retained and only the product code moved.

| mutation | expected | observed |
| --- | --- | --- |
| composer reverted to `dbeb1b3d` (#1607 out, #1606 in) | `openapi_v2` fails on the bodyless exception; `openapi_v3`, which declares none, still passes | `SwaggerV2Downgrader: response media types require a shared schema or example.` — `openapi_v2: failed`; `openapi_v3: 44,950 ms` pass |
| composer reverted to `42f9c5b6` (#1606 and #1607 both out) | `openapi_v3` fails on the `void` route's `content`, with nothing refusing the document | `Bug on void success carries no content: found different values - []: { "x": { "application/json": {} } ... }` — `openapi_v3: failed`, while the two adjacent cases in the same document, `test_openapi_v3` and `test_openapi_v3_default_server_keeps_description`, both stayed green |

The first is the one that matters for the shape of these tests: the mutation is targeted, so a control feature one property away stays green rather than the whole suite going red.

Restored and re-run at the final composer: `openapi_v2: 49,275 ms`, `openapi_v3: 49,487 ms`, both green.

## Integration-sensitive merge gate

## Integration-sensitive merge gate

The native Go transform, package manifests, the SDK generators and CLI-driven fixtures all changed this cycle, so the [issue-campaign rule](https://github.com/samchon/nestia/blob/master/.agents/skills/issue-campaign/development.md) requires constructing the prospective merge in a disposable checkout and running the canonical set there. This closes the gap the handoff comment named: a full `pnpm test` had been started against `eb8f8b8a` and never reached the test workspaces.

**SHAs.** `origin/master` `a11f7167` (re-read immediately before the gate and again before merge; unmoved) and head `68600ba7`. `origin/master` is an ancestor of the head, so the prospective merge tree is the head tree — verified with `git merge-base --is-ancestor`, not assumed.

**Where.** A disposable clone at `/d/nestia-gate-1599`, detached at the head, `git clean -xdf` outside `node_modules`.

**Runtime.** Node **v24.19.0**, the version every workflow pins. This is not incidental: the machine's default Node is 22.21.0, and on that runtime two `test-sdk` features fail with `request for './max.js' is from a module not been linked` at this head *and at `dbeb1b3d`*. That is #1609, filed from this gate, and it is why the gate runs on the contracted runtime rather than the one that happened to be on `PATH`.

## Commands and results

| step | command | result |
| --- | --- | --- |
| install | `pnpm install` | 0 |
| full suite | `pnpm test` — build, `test:go`, then all seven `tests/test-*` at `--workspace-concurrency=1` | 127 — 153 of 155 `test-sdk` features green, two red; every other workspace green. The two are #1611, below |
| packaging | `pnpm run package:tgz` | 0 |
| packed consumer | tarball inspection + clean `npm install` of all eight tarballs outside the workspace + entry-point resolution | 0 — all eight packages resolve |

## The packed consumer check, and why it belongs to this cycle

The `ttsc >=0.19.2` floor only means anything because the shipped package has no `go.work`: ttsc generates its own workspace from the installed npm ttsc's Go tree, and refuses a plugin `go.mod` that replaces a ttsc-managed module. If `go.work` shipped, the plugin would compile against this repository's pins on a consumer's machine and the floor would be describing something else entirely. So the check asserts it directly on the artifact rather than on `files`:

```
===== packed @nestia/core native tree =====
cmd  go.mod  go.sum  plugin  transform  transform.cjs
OK: no go.work in the tarball
OK: native/go.mod
OK: native/go.sum
OK: native/transform.cjs
peerDependencies.ttsc = >=0.19.2
peerDependenciesMeta.ttsc = {"optional":true}
version = 12.1.0
```

`version = 12.1.0` is the other thing worth reading there: unchanged, as [Change Integrity](https://github.com/samchon/nestia/blob/master/.agents/skills/development/SKILL.md#change-integrity) requires of an implementation pull request. The native tree changed this cycle, so it earns a version from a separate maintainer release commit rather than from here.

Then, from an empty project at `/d/nestia-consumer-check` outside the workspace, `npm install` of all eight tarballs (270 packages, exit 0) and a `require` of each public entry through that install's own resolver:

```
OK   @nestia/fetcher   -> 3 exports
OK   @nestia/core      -> 23 exports
OK   @nestia/sdk       -> 4 exports
OK   @nestia/e2e       -> 7 exports
OK   @nestia/migrate   -> 18 exports
OK   @nestia/benchmark -> 1 export
OK   nestia            -> 12.1.0, bin { "nestia": "./bin/index.js" }
OK   @nestia/editor    -> 12.1.0
OK   @nestia/core/native/transform.cjs -> function
```

The last line is the one this cycle put at risk: `transform.cjs` is the descriptor ttsc loads to find the Go source, and it has to resolve through the exports map from a consumer's `node_modules`, not from the workspace.

## What the gate found: #1611

The gate is the reason this pull request has a seventh commit. `test-sdk` ran 153 of 155 features green and failed two, `swagger` and `swagger-한글경로`, both before any assertion:

```
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node
are supported by the default ESM loader. On Windows, absolute paths must be valid file://
URLs. Received protocol 'd:'
    at load (ttsc/lib/launcher/internal/runtimeHooks.js:248:16)
```

The frame points into ttsc, but the cause is nestia's own fixture — one line, identical in both files:

```ts
const { NESTIA_CONFIG } = await import(__dirname + "/../../../../nestia.config.ts");
```

`import()` takes a URL specifier, not a filesystem path. On POSIX the two coincide for an absolute path, so it resolves on every runner CI uses; on Windows the path starts with a drive letter and Node reads `D:` as a protocol. The `fs.promises.readFile` calls three lines below take a path and are unaffected, which is why only the import fails.

Attribution did not need a build: `git diff a11f7167 68600ba7` over both files is empty and the offending line is present at the merge base, so this pull request neither introduced it nor can be blamed for it. It also reproduces on Node 22.21.0, so it is not #1609 either.

The differentiator is the import and nothing else. These two files are the only path-based dynamic imports anywhere under `tests/`, and they are exactly the two failures; `swagger-extensions` has the same case-file name, the same nesting under `src/test/features/swagger/`, and the same `ttsx -r` invocation, and passes — its version reads `swagger.json` with `fs` and imports nothing.

Fixed rather than explained away, because a local full `pnpm test` **is** this gate: `pathToFileURL(path.join(__dirname, ...)).href`, which also percent-encodes, as the `swagger-한글경로` path needs. The `fs` reads stay as they are; rewriting them would suggest the two APIs share a contract when the whole defect is that they do not.

This is also why `test-transform-options` shows separately below: `test-sdk` failing stops the recursive run before it, so it was executed on its own at the same head — `validate options`, `stringify options`, `llm strict diagnostics`, `disabled transform`, `transform envelope graph`, `aliased core imports`, exit 0.

## Pull-request CI at `68600ba7`

All twenty checks green at `68600ba7`, the same immutable head this gate ran against. That head is no longer the tip — the gate's own finding produced `d76e8d67` — so both CI and the gate re-run against the new head, and those results are the ones the merge decision uses. This matrix is recorded because it is what proved the cycle's substantive work green before the portability fix was added on top:

| lane | | lane | |
| --- | --- | --- | --- |
| `sdk` | pass 15m24s | `go` | pass 16m58s |
| `migrate` | pass 8m05s | `e2e` | pass 7m23s |
| `cli` | pass 7m47s | `editor` | pass 7m36s |
| `transform-options` | pass 7m51s | `benchmark` | pass 7m35s |
| `benchmark-system (ubuntu)` | pass 8m13s | `benchmark-system (windows)` | pass 10m09s |
| `Ubuntu` | pass 6m58s | `deploy` | pass 6m17s |
| `format` | pass 32s | `spell-check` | pass 5s |
| `Analyze (go)` | pass 2m17s | `Analyze (javascript)` | pass 1m12s |
| `Analyze (actions)` | pass 43s | `CodeQL` | pass |
| `Socket · project` | pass | `Socket · pull request` | pass |

Worth naming what CI does **not** cover, since the gate exists for exactly that: every workspace lane runs `ubuntu-latest`, and the only Windows lanes are `benchmark-system`, which builds and tests the benchmark workspace rather than any `tests/test-*`. So this gate is the first full `pnpm test` this branch has had on Windows, and #1609 and #1610 are both things it found there.

## A first attempt that does not count, recorded rather than dropped

The gate's first `pnpm test` reported 50 failed `test-sdk` features. It is not a result. Roughly thirty features passed and then every remaining one died with exit code `3221225794` — `0xC0000142`, `STATUS_DLL_INIT_FAILED`, which is Windows refusing to start a process rather than anything the process did. Two other `test-sdk` suites were running beside it at the time, for the mutation records above, and the three together exhausted the desktop heap.

It is recorded because a reader of the log would otherwise see fifty red features and conclude the pull request is broken. The re-run above is the gate, run alone on the machine.

What the aborted attempt did establish before it ran out of resources, and the re-run confirms: `pnpm build`, `pnpm test:go`, `test-benchmark`, `test-cli`, `test-e2e`, `test-editor` and `test-migrate` all green on Windows/Node 24.

## Closing keywords, reconciled against what survives at `HEAD`

`git log origin/master..HEAD` carries ten closing lines. Nine have a surviving fix at `HEAD`:

| issue | surviving fix |
| --- | --- |
| #1598 | the `graph` section in `runTransformProject`, plus the `ttsc` peer floor |
| #1600 | `.prettierignore`, `format:check`, `format.yml` |
| #1602 | the skill counts, corrected again in `cd02e681` for the tree the revert left |
| #1603 | `bundle.js` resolving a plain specifier against both catalogs |
| #1605 | the default server's description dropped for the one target that cannot carry it |
| #1606 | `content` omitted when the success response has no body |
| #1607 | the same rule for a declared exception, through one shared helper |
| #1608 | the `openapi` option's value set and default |
| #1611 | `pathToFileURL` in the two swagger fixtures |

**#1601 is the tenth and has no surviving fix.** `679355e8` carries `Close #1601` and `897af4ab` reverted it. This repository merges cycle pull requests with a merge commit — #1591 and #1593 both did — so every commit message lands on `master` and GitHub will close #1601 mechanically. That is a property of the merge, not a decision, and the merge gate reopens it immediately afterwards with a comment saying so. It stays blocked on #1604.

## What this cycle leaves open, deliberately

| issue | why it is not in this pull request |
| --- | --- |
| #1601 | needs the typia 13.2.0 pin, which `897af4ab` reverted because that release renames a duplicated metadata separator the SDK clone generator reads; blocked on #1604 |
| #1604 | fixing it renames published generated output, which is a maintainer's product decision rather than a side effect of a cache-invalidation cycle |
| #1611 | **fixed here**, in the commit the gate produced. It is the one thing on this list that is not deferred: a local full `pnpm test` is the gate itself, so leaving it red would mean merging on a gate that cannot pass |
| #1610 | `format:check`, which this cycle adds, is red on an unmodified Windows checkout because Prettier forces LF and nothing normalizes the working tree. Two remedies exist with different costs -- a tolerant checker or a `.gitattributes` that makes every checkout uniform -- and picking one silently inside this cycle would settle a repository-wide policy by side effect |
| #1609 | found by this gate; choosing the repository's Node floor is a policy decision, and the evidence here establishes only that two `test-sdk` features fail on 22.21.0 |
| #1598 step 3 | `dependenciesComplete` is not declarable by this host, with the reason recorded at the fill site; unblocking needs samchon/typia#2357 and the untouched-file rule from samchon/ttsc#1259 |
| `requestBody` | `SwaggerOperationParameterComposer.body` is the last site that builds a media type by hand. Untouched by this pull request and no more reachable than before; whether `@TypedBody() input: void` survives the transform's diagnostics was not established, and a fix without that answer would be a guess |

samchon (commented):
# Self-Review round 4 — fresh pass over `origin/master...d76e8d67`

The gate produced `d76e8d67`, so this is the complete round after it. 59 changed files now, all read again at this head. **No finding survived verification**, and no work followed, so this is the clean round that ends the review.

## The one hunk this round had not seen

`d76e8d67` itself. Reviewing it turned up nothing to change, but one check it prompted is worth recording, because the [Forbidden](https://github.com/samchon/nestia/blob/master/.agents/skills/development/SKILL.md#forbidden) section's fourth rule is exactly about this: a defect found in two files is a class until proven otherwise.

Swept every dynamic `import()` in `packages`, `tests`, `benchmark`, `deploy` and `config`. There are three in the repository:

| site | specifier | portable |
| --- | --- | --- |
| `tests/.../swagger/.../test_swagger.ts` | was `__dirname + "/..."` | fixed |
| `tests/.../swagger-한글경로/.../test_swagger.ts` | was `__dirname + "/..."` | fixed |
| `packages/sdk/src/executable/sdk.ts` | `"./internal/NestiaSdkCommand"` | **yes, and untouched** |

The third one is production code in a published CLI, so it was the one worth being wrong about. It is a *relative* specifier, resolved against the importing module's own URL, which is portable on every platform — not an instance of this defect, and it stays as it is. With that, the class is closed rather than the two witnesses patched.

## Re-derived at this head

- 59 files reconciled against `git diff --name-status`; nothing unexamined
- every documented count re-counted from `git ls-files` and holding: `packages/core/test` 61, `packages/sdk/test` 19, tracked `*_test.go` 80, `transform/` 13, `test-sdk` features 155, packages 8, test workspaces 7, `replace` directives 15 of which 14 are shim redirects
- `0` `version` fields anywhere in the diff
- `gofmt -l` over all 100 tracked `*.go` blobs — clean
- `pnpm run format:check` — clean

## A judgment call, recorded rather than silently taken

`test_swagger.ts` carries no doc comment, and the [testing convention](https://github.com/samchon/nestia/blob/master/.agents/skills/development/SKILL.md#testing) asks for the three-part shape on every *new or materially changed* case. This change alters how the module is loaded and nothing about what the case asserts — `info.version`, `info.title`, `info.description` and `servers` are untouched — so it is not a material change to the case, and back-filling a doc comment on a pre-existing undocumented fixture would widen a portability fix into a documentation pass. Left alone deliberately.

## Verification at this round

- `pnpm --filter ./tests/test-sdk start -- --only swagger` on Windows / Node 24: all seven features green, including the two that were red — `swagger-한글경로` 78,335 ms and `swagger` 78,455 ms

### The gate, re-run whole at this head

Not a partial re-run of the two features that were red. The same disposable clone, moved to `d76e8d67` (tree `23493117`, matching this checkout exactly), against an `origin/master` still at `a11f7167`:

| step | result |
| --- | --- |
| `pnpm install` | 0 |
| `pnpm test` — build, `test:go`, all seven `tests/test-*` at `--workspace-concurrency=1` | **0** |
| `pnpm run package:tgz` | 0 |

All seven workspaces ran to completion this time, `test-transform-options` included — it was unreachable before, because `test-sdk` failing stops the recursive run ahead of it. Its six cases: `validate options` 13,814 ms, `stringify options` 7,715 ms, `llm strict diagnostics` 5,566 ms, `disabled transform` 624 ms, `transform envelope graph` 1,725 ms, `aliased core imports` 1,305 ms.

`test-sdk` accounts for all 155 features: 113 emit a timing, and the other 42 are exactly the `error` and `distribute` families, whose expectation is encoded in the directory name rather than in a duration — an `error` feature *must* fail to compile and the harness throws if it succeeds, and a `distribute` feature returns after its output is cleaned. Reconciled against `ls tests/test-sdk/features` rather than eyeballed: 113 + 42 = 155, nothing unaccounted for, and zero failures.

Both Go modules: `ok github.com/samchon/nestia/packages/core/test`, `ok github.com/samchon/nestia/packages/sdk/test`.

### Pull-request CI at this head

All twenty green at `d76e8d67`:

| lane | | lane | |
| --- | --- | --- | --- |
| `sdk` | pass 18m36s | `go` | pass 18m31s |
| `benchmark-system (windows)` | pass 11m32s | `deploy` | pass 9m28s |
| `benchmark` | pass 8m01s | `migrate` | pass 7m57s |
| `editor` | pass 7m47s | `benchmark-system (ubuntu)` | pass 7m44s |
| `cli` | pass 7m39s | `Ubuntu` | pass 7m35s |
| `transform-options` | pass 7m20s | `e2e` | pass 7m18s |
| `Analyze (go)` | pass 2m17s | `Analyze (javascript)` | pass 1m07s |
| `Analyze (actions)` | pass 47s | `format` | pass 30s |
| `spell-check` | pass 7s | `CodeQL` | pass |
| `Socket · project` | pass | `Socket · pull request` | pass |

The `sdk` lane is the one that matters for this cycle's substance, and it is the Linux counterpart of the gate above: same 155 features, same generators, different platform.

### Packaging, and why the packed consumer is not re-run

`pnpm run package:tgz` runs again at this head. The clean packed-consumer install is not repeated, and that is a reasoned omission rather than a skipped step: `git diff 68600ba7 d76e8d67` is exactly the two files under `tests/`, and `tests/` appears in no package's `files` array, so every tarball is identical to the one the consumer check already installed and resolved. Re-running it would re-verify bytes that did not move.

## Merge decision

Every precondition the [campaign's merge gate](https://github.com/samchon/nestia/blob/master/.agents/skills/issue-campaign/development.md) names is now satisfied against one immutable head, `d76e8d67`:

- required checks green at that head, not at an older one
- a complete Self-Review round with no sound improvement, after the last correction rather than before it
- the integration gate run on the prospective merge of an unmoved `origin/master` — re-read again immediately before merging — with packaging and a clean packed consumer
- every product change in the cycle proved sensitive by mutation
- closing keywords reconciled against what survives at `HEAD`, with the one that does not — #1601 — named in advance and reopened immediately after

Merging with a merge commit, which is what this repository does with cycle pull requests (#1591, #1593); the dependabot squashes are the exception, not the rule. No `version` field moves: the native tree changed, so it earns one from a separate maintainer release commit.

Issue #1609, opened 2026-08-20
the repository declares no Node version, and below the one CI uses two test-sdk features fail with an error that names neither


samchon:
## Problem

The repository declares no Node version anywhere a contributor's tooling can read it. There is no `engines` field in the root `package.json` or in any of the eight packages, and no `.nvmrc`. The only statement of the supported runtime is one clause in `.agents/skills/project/SKILL.md` — "the `NODE_OPTIONS=...` that Node 24 needs" — plus `node-version: 24.x` repeated in every workflow.

Below that version the failure is not a warning and does not name Node. Two `test-sdk` features die during generation with:

```
request for './max.js' is from a module not been linked
```

which points at neither the runtime nor the package responsible, and `pnpm install` accepts the environment without a word.

## Evidence

Reproduced on Node **v22.21.0**, Windows 11, at two commits — `dbeb1b3d` and the head of #1599 — so it is a property of the environment rather than of any pending change:

```
$ TEST_SDK_SKIP_BUILD=1 pnpm --filter ./tests/test-sdk start -- --only swagger-example
Test Features (concurrency: 1)
 Nestia All Generator
request for './max.js' is from a module not been linked
  - swagger-example: failed
```

Same for `exception-filter`. Deterministic at concurrency 1, so it is not a parallelism artifact.

The two failing features are exactly the `test-sdk` features that import `uuid`:

| feature | imports `uuid` | result on Node 22 |
| --- | --- | --- |
| `swagger-example` | yes | fails |
| `exception-filter` | yes | fails |
| `exception` | no | passes |
| `openapi_v2` | no | passes |
| `swagger-customizer` | no | passes |
| `status` | no | passes |

The same feature, same checkout, same command, with only Node swapped to **v24.19.0**:

```
$ TEST_SDK_SKIP_BUILD=1 pnpm --filter ./tests/test-sdk start -- --only swagger-example
  - swagger-example: 66,845 ms
Total Elapsed Time: 66,866 ms
```

`uuid@13.0.2` ships `dist/max.js`, and the failure is Node's module customization hooks (`node:internal/modules/customization_hooks`, reached through `ttsc/lib/launcher/internal/runtimeHooks.js`) failing to link that specifier. The synchronous hooks API those runtime hooks use is the part of Node that moved between 22 and 24, which is what makes the version the variable.

[...]

Issue #1604, opened 2026-08-19
the SDK clone generator reads a typia metadata name as a dot-separated identifier chain, so typia 13.2's duplicate-name id emits code that does not parse


samchon:
## Problem

`@nestia/sdk`'s clone generator treats a typia metadata name as a dot-separated chain of TypeScript identifiers. It is not one. typia mints a disambiguating id for a duplicated name, and since 13.2.0 that id is separated by `-o` instead of `.`, so the generator emits a type name and an import binding that do not parse.

This blocks every future move of the `typia` pin in `packages/*/native/go.mod`. It is not a hypothetical: it was reproduced by moving that pin and reverted in #1599 rather than fixed, because fixing it means choosing what a generated SDK calls a disambiguated type, which is a change to public generated output.

## Evidence

Move the Go `typia` pin from `v0.0.0-20260709134626-fd1c6686950f` (between 13.0.0 and 13.0.1) to `v0.0.0-20260722004126-436fd9275bb5` (13.2.0) in `packages/core/native/go.mod`, `packages/sdk/native/go.mod`, `packages/core/test/go.mod` and `packages/sdk/test/go.mod`, then:

```bash
pnpm --filter ./tests/test-sdk start
```

Four features fail — `clone-and-keyword`, `clone-and-propagate`, `clone-and-propagate-and-keyword`, `clone-implicit` — all at the compile step, on generated SDK source:

```
src/api/functional/arrayRecursiveUnionImplicit/index.ts:13:25 - error TS1005: ',' expected.

13 import type { IDirectory-o1 } from "../../structures/IDirectory-o1";
                           ~
```

`IDirectory-o1` is not an identifier. `clone`, `clone-native`, `clone-tags` and `clone-type-create-duplicate` pass, so the trigger is a duplicated name, not cloning as such.

## Root cause

typia's `MetadataCollection.getName` mints an id for a name it has already handed out, and 13.2.0 changed the separator deliberately. From `native/core/schemas/metadata/MetadataCollection.go`:

```go
// metadataCollection_duplicateSuffix separates a duplicated base name from its
// disambiguating counter.
//
// It must not be ".", which is how a qualified name separates a namespace from
// its member. Two consumers read that meaning back out of an allocated id: ...
// A dotted counter also let an invented id land on the real full name of a
// `namespace Foo { interface o1 }` member, and the loser was dropped from the
// document outright.
const metadataCollection_duplicateSuffix = "-o"
```

[...]

Issue #1601, opened 2026-08-19
the nestia transform host never registers typia default-library classifier, so a user-authored global is validated as a runtime native


samchon:
## Problem

`@nestia/core` hosts typia's transform instead of reimplementing it, but it does not install the per-program registration typia's own host installs before that transform runs. typia's metadata analysis therefore decides **runtime-native identity** from a file-name pattern rather than from the program's actual default-library set.

The consequence is a wrong validator, not a slower one. A purely user-authored global whose name matches a runtime native is promoted to native identity: its members stop being validated and an `instanceof` check takes their place, against a constructor that need not exist at runtime.

## Evidence

Reproduce with two files in any project compiled through `@nestia/core`'s transform. No `@nestia/core` import is required — the host runs typia's transformer over every source file:

```ts
// src/lib.custom.d.ts   (a project file, not a default library)
interface Blob {
  customField: string;
}
```

```ts
// src/main.ts
import typia from "typia";

export interface IHolder {
  blob: Blob;
}

export const check = typia.createIs<IHolder>();
```

with a tsconfig that leaves the program's only `Blob` to that declaration file:

```jsonc
{
  "compilerOptions": {
    "lib": ["ESNext"],
    "types": [],
    "plugins": [{ "transform": "@nestia/core/native/transform.cjs" }]
  },
  "include": ["src"]
}
```

Transformed output, read from `TtscCompiler.transform()`:

```ts
export const check = (() => {
    const _io0 = (input: any): boolean => input.blob instanceof Blob;
    return (input: any): input is IHolder => "object" === typeof input && null !== input && _io0(input);
})();
```

`customField` is never checked, and `instanceof Blob` throws `ReferenceError: Blob is not defined` on any runtime that does not provide the constructor — which is the whole premise of the fixture, since nothing in the program declares `Blob` as a runtime authority.

The expected output, which the same program produces once the classification comes from the program, is the structural one:

```ts
const _io0 = (input: any): boolean => "object" === typeof input.blob && null !== input.blob && _io1(input.blob);
const _io1 = (input: any): boolean => "string" === typeof input.customField;
```

## Root cause and ownership

The decision lives in typia and is documented there. `native/core/factories/internal/metadata/MetadataHelper.go`:

```go
func metadata_declaration_is_runtime_provided(checker *nativechecker.Checker, declaration *nativeast.Node) bool {
[...]

Issue #1600, opened 2026-08-19
pnpm format rewrites two generated trees, and nothing checks the tracked ones, so 21 source files drifted


samchon:
## Problem

`pnpm format` is the repository's only formatting gate and nothing verifies it ran. Twenty-one tracked TypeScript files are currently not formatter-clean, so the next contributor who follows the documented rule — [`.agents/skills/development/SKILL.md`](https://github.com/samchon/nestia/blob/master/.agents/skills/development/SKILL.md): *"Run `pnpm format` before every ordinary commit and stage the result; never commit unformatted output"* — gets twenty-one unrelated files in their diff and has to choose between shipping the noise and disobeying the rule.

The same run also rewrites two generated trees that no one should be formatting, and both are regenerated by the very next `pnpm install`, so the rewrite is pure churn that reverts itself.

## Evidence

Reproduce from a clean checkout of `master`:

```bash
pnpm install
pnpm format
git diff --name-only
```

Twenty-one tracked files change. They are ordinary source and test files across six packages and six test workspaces:

| area | files |
| --- | --- |
| `packages/benchmark` | `src/DynamicBenchmarker.ts` |
| `packages/core` | `src/decorators/McpRoute.ts`, `src/decorators/SwaggerCustomizer.ts`, `src/decorators/SwaggerExample.ts` |
| `packages/e2e` | `src/GaffComparator.ts` |
| `packages/sdk` | `src/analyses/ImportAnalyzer.ts` |
| `tests/test-benchmark` | `src/index.ts` |
| `tests/test-cli` | `src/features/test_cli_package_manager_missing_halt.ts`, `test_cli_repository_option_override.ts`, `test_cli_start_command_sequence.ts`, `test_cli_template_command_sequence.ts` |
| `tests/test-e2e` | `src/features/test_gaff_comparator_prefix_keys.ts` |
| `tests/test-editor` | `src/features/test_editor_uploader_default_package.ts` |
| `tests/test-migrate` | `src/features/test_migrate_sdk_empty_paths.ts` |
[...]

PR #1612, opened 2026-08-20
fix(repo): normalize checkouts to LF so format:check is platform-neutral

### LINKED ISSUE #1610: format:check is red on an unmodified Windows checkout, because Prettier forces LF and nothing normalizes the working tree
## Problem

`prettier.config.js` leaves `endOfLine` unset, so Prettier uses its default of `"lf"`. The repository carries no `.gitattributes`, so on Windows a checkout with the default `core.autocrlf=true` puts CRLF in every working file. `pnpm run format:check` — added in #1599 and now gating every pull request — therefore reports every unchanged file as a style violation on a Windows checkout, and `pnpm format`, which the development skill requires before every ordinary commit, rewrites hundreds of files that produce no commit.

The loop is stable in both directions and neither end is a real formatting change:

1. `pnpm run format:check` → "Code style issues found in N files"
2. `pnpm format` → rewrites them to LF; `git diff` is empty because git normalizes on commit
3. any `git checkout -- <path>` writes CRLF back
4. return to 1

## Evidence

Windows 11, `core.autocrlf=true`, at `68600ba7`:

```
$ git diff HEAD --stat          # working tree identical to HEAD
$ npx prettier --check "packages/**/*.ts" "tests/**/*.ts"
[warn] packages/sdk/src/analyses/ImportAnalyzer.ts
... 27 more ...
[warn] Code style issues found in 28 files. Run Prettier with --write to fix.
```

The 28 files are exactly the contents of one directory that had just been through `git checkout <sha> -- packages/sdk/src`; nothing else distinguishes them, and their committed blobs are byte-identical to `HEAD`.

CI never sees it: the `format` lane runs `ubuntu-latest`, where the checkout is LF, and it is green at this commit. Only `benchmark-system` runs on Windows, and it does not run the formatter.

Before #1599 the repository had no `format:check`, so this cost a contributor nothing beyond a noisy `git status` after the documented pre-commit `pnpm format`. With the check gating pull requests it now looks like a red gate on a clean tree.

## Consequence surface

| surface | effect |
| --- | --- |
| a Windows contributor | `pnpm run format:check` is red on an unmodified checkout, and `pnpm format` rewrites files that produce no diff |
| the documented pre-commit step | `pnpm format` is required before every ordinary commit, so this is hit on the first commit rather than as an edge case |
| the `format` CI lane | unaffected; it is green and correct on Linux |
| the committed tree | unaffected; git normalizes to LF on commit, so no wrong bytes ever land |

## Approach

Two fixes exist and they are not equivalent, which is why this is filed rather than patched.

**`endOfLine: "auto"` in `prettier.config.js`.** One line. `--write` then preserves each file's existing endings and `--check` accepts either, so the Windows loop disappears. The cost is that the gate stops being able to reject a genuinely CRLF-committed file — reachable for a contributor with `core.autocrlf=false`, since nothing else in the repository normalizes.

**A `.gitattributes` with `* text=auto eol=lf`.** Makes every checkout LF regardless of a contributor's git configuration, so the current `endOfLine` default becomes correct everywhere and the gate keeps its full strength. The cost is that it changes the working copy of every existing Windows clone on the next checkout, and it is a repository-wide policy rather than a formatter setting.

Choosing between "make the checker tolerant" and "make the checkout uniform" is the maintainer's call. Whichever lands, the [Commands section](https://github.com/samchon/nestia/blob/master/.agents/skills/project/SKILL.md#commands) should say which one holds, so the next Windows contributor does not re-derive this.

## Acceptance and verification

- `pnpm run format:check` passes on an unmodified checkout on Windows and on Linux.
- `pnpm format` on an unmodified Windows checkout rewrites nothing.
- The `format` lane stays green, and a deliberately mis-formatted file still fails it from both the `packages/**` and `tests/**` globs.

## Coordination

Found in #1599's Self-Review, on the gate that same pull request adds. Not fixed there: both remedies are repository-wide decisions with different costs, and picking one silently inside a cache-invalidation cycle would settle a policy question by side effect. Same family as #1609 — the development environment is contracted only by what CI happens to run.

samchon:
## Intent

Close #1610: `pnpm run format:check`, which has gated every pull request since #1600, is red on an unmodified Windows checkout.

`prettier.config.js` leaves `endOfLine` at its `"lf"` default, so the formatter writes LF. Nothing normalized the working tree and the repository carried no `.gitattributes`, so a Windows clone with the default `core.autocrlf=true` gets CRLF on checkout. The check then reports every unchanged file as a style violation, and `pnpm format` — required before every ordinary commit — rewrites hundreds of files that produce no diff. The loop is stable in both directions: format, then any `git checkout -- <path>` writes CRLF back.

## Scope

One file. `.gitattributes` with `* text=auto eol=lf`, plus explicit `binary` for the tracked binary formats.

The alternative, `endOfLine: "auto"` in `prettier.config.js`, is one line shorter and was rejected: it makes the checker tolerant instead of the checkout uniform, so the gate would lose its ability to reject a genuinely CRLF-committed file — reachable for a contributor with `core.autocrlf=false`, since nothing else in the repository normalizes.

**No committed content moves.** Every text blob in the object store is already LF, verified by scanning `git show HEAD:<file>` for CR across every tracked `*.ts`, `*.js`, `*.json`, `*.md`, `*.mdx`, `*.yaml`, `*.yml`, `*.go`, `*.cjs` and `*.mjs` — zero hits. So `eol=lf` changes only what lands in a working tree, and no renormalization commit is needed.

## Verification

Windows 11, `core.autocrlf=true`, after `git rm --cached -r . && git reset --hard` to re-materialize the working tree under the new attributes:

| check | before | after |
| --- | --- | --- |
| working tree line endings | CRLF | LF |
| `pnpm run format:check` on an unmodified checkout | 28 files reported | `All matched files use Prettier code style!` |
| `pnpm format` on an unmodified checkout | rewrote hundreds of files | working tree unchanged, `git status` empty |

The gate keeps its strength — a deliberate violation still fails it from each glob separately, which is #1600's own acceptance criterion:

```
packages/** glob:  [warn] packages/core/src/decorators/TypedBody.ts
tests/** glob:     [warn] tests/test-e2e/src/features/test_validate_index.ts
```

Both restored afterwards; `format:check` clean and `git status` empty.

## Not in this pull request

#1609 is closed as not planned rather than fixed — the evidence only covered Node 22.21.0 and 24.19.0, so any `engines` floor written from it would exclude untested versions.

#1604 and #1601 stay open, and the catalog is not the lever. Reading typia's Go source across releases: the `-o` duplicate separator that breaks `@nestia/sdk`'s dot-splitting arrives in **13.1.19**, while `MetadataDefaultLibrary_register`, which #1601 needs, arrives in **13.2.0**. There is no version that gives one without the other, so #1604 has to be fixed in `@nestia/sdk` first.

| typia | `-o` separator | `MetadataDependency_listen` | `MetadataDefaultLibrary_register` |
| --- | --- | --- | --- |
| 13.0.2 / 13.1.0 / 13.1.1 | no | no | no |
| 13.1.19 | **yes** | yes | no |
| 13.2.0 / 13.3.0 | yes | yes | **yes** |

https://claude.ai/code/session_01UpgzpPfMkrt8W6rAoofbkW

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,21 @@
+# Normalize line endings so every checkout is LF, on every platform.
+#
+# Prettier's `endOfLine` defaults to "lf" and `prettier.config.js` leaves it
+# there, so the formatter writes LF. Nothing else normalized the working tree,
+# which meant a Windows clone with the default `core.autocrlf=true` got CRLF on
+# checkout and `pnpm run format:check` -- gating every pull request since #1600
+# -- reported every unchanged file as a style violation, while `pnpm format`
+# rewrote hundreds of files that produced no commit.
+#
+# `eol=lf` fixes that at the checkout rather than by loosening the checker, so
+# the gate keeps its ability to reject a genuinely CRLF-committed file. Every
+# blob already in this repository's object store is LF, so this changes no
+# committed content -- only what lands in a working tree.
+* text=auto eol=lf
+
+# The two binary formats this repository actually tracks. `text=auto` already
+# detects both, so these only make the intent explicit and keep `git diff` from
+# rendering them as text. Deliberately not listing formats nothing tracks --
+# `*.svg` is XML and belongs with the text files above.
+*.png binary
+*.ico binary
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -10,6 +10,10 @@ on:
       - '.github/workflows/format.yml'
       - '.prettierignore'
       - 'prettier.config.js'
+      # `.gitattributes` decides the line endings a checkout gets, and
+      # Prettier's `endOfLine` compares against them, so it can turn this
+      # check red on its own without any `*.ts` file changing.
+      - '.gitattributes'
       - 'packages/**/*.ts'
       - 'tests/**/*.ts'
       - 'pnpm-lock.yaml'
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -25,7 +25,17 @@ permissions:
 
 jobs:
   format:
-    runs-on: ubuntu-latest
+    name: format (${{ matrix.os }})
+    # Both platforms, because the thing this gate can get wrong is
+    # platform-specific. Prettier compares against the line endings the
+    # checkout produced, so a formatting rule can hold on Linux and fail on
+    # Windows -- which is exactly what #1610 was, undetected for as long as
+    # this workflow ran on ubuntu alone.
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4


samchon (commented):
# Self-Review — four rounds over `origin/master...HEAD`

The pull request opened as one file and is now three commits. Both additions came from reviewing it, not from the original plan.

## Round 1 — consequence surface of `* text=auto eol=lf`

Everything a blanket EOL rule can break, checked rather than assumed:

| question | answer |
| --- | --- |
| Does any tracked file *require* CRLF? | No `.bat` or `.cmd` is tracked. The one shell script, `deploy/README.bash`, requires LF — so `eol=lf` removes a latent hazard rather than creating one, since a CRLF checkout can break a shell script's shebang. |
| Does any code depend on reading CRLF? | Three sites reference `\r\n`: both `FilePrinter.ts` and `tests/test-sdk/start.js`, all three of the form `.split("\r\n").join("\n")`. They *normalize* CRLF away, so under `eol=lf` they simply stop encountering it. `bundle.js` uses tolerant `\r?\n` regexes. Nothing requires it. |
| Is any tracked blob CRLF, needing a renormalization commit? | No. Scanned `git show HEAD:<file>` for CR across every tracked `*.ts`, `*.js`, `*.json`, `*.md`, `*.mdx`, `*.yaml`, `*.yml`, `*.go`, `*.cjs`, `*.mjs` — zero hits. So the object store is already LF and this changes only working trees. |
| Do the binary declarations match reality? | **No — finding, applied.** |

**Finding: three of the five `binary` patterns were speculative.** `*.png` (19 tracked) and `*.ico` (1) exist; `*.tgz`, `*.woff` and `*.woff2` are tracked nowhere. Config written for hypothetical files is config nobody can verify, so the list is now what the repository actually contains. `*.svg` is XML and deliberately stays with the text files.

## Round 2 — clean

Re-derived after the fix: 0 of **2421** tracked text files are CRLF in the working tree after re-materializing it under the new attributes, `format:check` clean, working tree empty.

## Round 3 — the gate did not run on the file that governs it

**Finding, applied in `20bed80a`.** `format.yml` carries a `paths:` filter and `.gitattributes` was not in it. But `.gitattributes` decides the line endings a checkout gets, and Prettier's `endOfLine` compares against exactly that — so it can turn `format:check` red on its own with no `*.ts` file changing.

This pull request was its own witness: before that commit, the checks that ran were Analyze, CodeQL, Socket and spell-check. `format`, the gate this change exists to fix, was skipped.

The workflow's own header comment already records this reasoning for the `tests/**` glob. It just had not been applied to the new dependency.

## Round 4 — the fix had no regression guard

**Finding, applied in `12a15a0e`.** Nothing would notice if `.gitattributes` were removed or weakened later.

The reason is worth stating plainly, because it also explains why the CI on this pull request cannot by itself prove the fix: **on Linux the whole question is invisible.** `core.autocrlf` defaults to false there, checkouts were already LF, and `eol=lf` changes nothing — which is precisely why this lane stayed green for the entire life of #1610. The evidence for the fix is the Windows verification below, not a green Linux lane.

So the gate now runs on both platforms, the same `fail-fast: false` matrix shape `benchmark-system` and `test.yml` already use. `format:check` was added in #1600 to stop formatting drift, and the one platform where formatting behaves differently was the one it never ran on.

One thing I did **not** verify: whether the `windows-latest` runner image leaves `core.autocrlf` at Git for Windows' `true` default. If it does, removing `.gitattributes` turns that lane red, and the guard is real. If the image sets it false, the lane still exercises Windows but would not catch that particular removal. Recorded rather than claimed either way. Checking the job's name change is safe regardless: `master` carries no branch protection, so renaming the check from `format` to `format (ubuntu-latest)` breaks no required-check configuration.

## Verification

Windows 11, `core.autocrlf=true`, after `git rm --cached -r . && git reset --hard` to re-materialize the tree under the new attributes:

| | before | after |
| --- | --- | --- |
| tracked text files that are CRLF | all of them | **0 of 2421** |
| `format:check` on an unmodified checkout | 28 files reported | `All matched files use Prettier code style!` |
| `pnpm format` on an unmodified checkout | rewrote hundreds of files | working tree unchanged, `git status` empty |

The gate keeps the strength #1600 gave it — a deliberate violation still fails it from each glob separately:

```
packages/**  ->  [warn] packages/core/src/decorators/TypedBody.ts
tests/**     ->  [warn] tests/test-e2e/src/features/test_validate_index.ts
```

Both restored; `format:check` clean and `git status` empty afterwards.

`.github/workflows/format.yml` parses, and its `paths` resolves to the nine entries intended with `.gitattributes` among them.

PR #1625, opened 2026-08-22
fix(core): validate transforms without emit

### LINKED ISSUE #1624: @nestia/core noEmit skips transform diagnostics and failed builds publish output
## Summary

`@nestia/core` exits successfully before running its typia/core/SDK contributor transforms when a project uses `--noEmit`, configured `noEmit`, or the native `check` command. This makes the runtime `NoTransformConfigurationError` advice to run `npx ttsc --noEmit` ineffective for transform-only failures.

An emitting build has a related recovery problem: it writes declarations and JavaScript before checking accumulated transform diagnostics. The command returns exit code 3, but leaves a bare, untransformed `TypedRoute` decorator that can throw `NoTransformConfigurationError` if the failed output is executed.

## Reproduction

Compile a `TypedRoute` with LLM validation enabled and a response type that JSON accepts but LLM schemas reject:

```ts
interface IResponse {
  pair: [string, number];
}

class Controller {
  @TypedRoute.Get()
  public get(): IResponse {
    return { pair: ["one", 1] };
  }
}
```

- A normal `ttsc` build returns exit code 3 with `LLM schema does not support tuple type`, but currently leaves emitted files.
- `ttsc --noEmit` and the native `check` route currently return exit code 0 with no transform diagnostic.

## Expected behavior

- All no-emit routes run the same transform validation and report the exact `nestia.core.TypedRoute` diagnostic without publishing artifacts.
- An emitting build publishes files and its manifest only after TypeScript, emit, and transform diagnostics are clean.
- Analysis-only compiler options such as `allowImportingTsExtensions` remain valid during the private transform traversal.

This is the nestia aggregate-host counterpart to samchon/typia#2373 and samchon/typia#2374.

samchon:
## Intent

Make `ttsc --noEmit`, configured `noEmit`, and the native `check` command surface the same typia/core/SDK contributor diagnostics as an emitting nestia build, while ensuring failed builds cannot leave runnable untransformed output.

Closes #1624.

## Scope

- reload a private emit-enabled program after the original no-emit program passes TypeScript diagnostics, then run the complete composed transform pipeline with every output discarded;
- route `check` through the same no-emit build path so the two commands cannot diverge;
- buffer declarations, JavaScript, maps, and build metadata until emit and transform diagnostics are clean;
- pin LLM-only tuple return diagnostics for `TypedRoute` across check, explicit `--noEmit`, configured `noEmit`, and normal failed emit paths;
- assert the real `ttsc` CLI reports the detailed `nestia.core.TypedRoute` diagnostic and publishes no failed output;
- update typia, `@typia/interface`, and `@typia/utils` to 14.0.2 so nestia consumes the upstream runtime guidance release.

`@nestia/sdk` needs no separate execution change: its compiler already propagates the core host's nonzero status and stderr. No package version fields are changed; release versioning remains maintainer-owned.

## Local verification

- `go test ./... -run 'TestBuild(NoEmitReportsLlmRouteDiagnostic|TransformDiagnosticPublishesNoArtifacts|NoEmitPreservesAnalysisOnlyOptions)$' -count=1 -v`
- `pnpm --filter ./tests/test-transform-options start`
- `pnpm format`
- `git diff --cached --check`

The full `pnpm --filter @nestia/core test:go` run was intentionally stopped before completion at maintainer direction; the complete repository matrix is deferred to GitHub Actions.

### DIFF
--- a/packages/core/native/transform/build.go
+++ b/packages/core/native/transform/build.go
@@ -68,23 +68,54 @@ func runBuild(args []string) int {
 		return 2
 	}
 	if len(diags) > 0 {
+		if prog != nil {
+			prog.Close()
+		}
 		driver.WritePrettyDiagnostics(stderr, diags, cwd)
 		return 2
 	}
-	defer prog.Close()
-	releaseTypiaRegistries := registerTypiaDefaultLibraryClassifier(prog)
-	defer releaseTypiaRegistries()
 	if profile {
 		started = time.Now()
 	}
 	if diags := prog.Diagnostics(); len(diags) > 0 {
 		profileBuildStep(profile, "diagnostics", started)
+		prog.Close()
 		driver.WritePrettyDiagnostics(stderr, diags, cwd)
 		return 2
 	}
 	profileBuildStep(profile, "diagnostics", started)
 
 	shouldEmit := !prog.ParsedConfig.ParsedConfig.CompilerOptions.NoEmit.IsTrue()
+	if !shouldEmit {
+		// Plugin transformers are part of tsgo's emit pipeline. Reload with emit
+		// enabled so check/noEmit traverses the same source set, then discard every
+		// generated output below. The original program already proved the project's
+		// diagnostics; diagnosing this private reload would reject valid analysis-
+		// only options such as allowImportingTsExtensions.
+		prog.Close()
+		if profile {
+			started = time.Now()
+		}
+		prog, diags, err = driver.LoadProgram(cwd, *tsconfigPath, driver.LoadProgramOptions{
+			ForceEmit: true,
+			OutDir:    *outDir,
+		})
+		profileBuildStep(profile, "load-transform-program", started)
+		if err != nil {
+			fmt.Fprintf(stderr, "ttsc-nestia build: %v\n", err)
+			return 2
+		}
+		if len(diags) > 0 {
+			if prog != nil {
+				prog.Close()
+			}
+			driver.WritePrettyDiagnostics(stderr, diags, cwd)
+			return 2
+		}
+	}
+	defer prog.Close()
+	releaseTypiaRegistries := registerTypiaDefaultLibraryClassifier(prog)
+	defer releaseTypiaRegistries()
 	if !*quiet {
 		fmt.Fprintf(
 			stdout,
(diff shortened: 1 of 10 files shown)


samchon (commented):
## Self-Review round 1 — finding fixed

Reviewed immutable commit `a4339e8c3ea36ca53854c6bfa1eff311ddb3a750` across the complete pull-request diff and the compiler-driver, SDK consumer, platform, recovery, packaging, documentation, and CI consequence surfaces.

One sound concurrency defect survived verification: TypeScript-Go's declaration emitter queues source files in parallel, while the new deferred-output callback appended to the shared `emitted` and `pending` slices without synchronization. A declaration build could therefore race or lose a buffered artifact. Commit `cdd8d09af8aabee9320e632da2bcd44336658b31` serializes those callback writes.

Because the review produced a change, this round is not the final clean gate. A fresh whole-surface round is running against `cdd8d09af8aabee9320e632da2bcd44336658b31` while its replacement CI matrix executes.

samchon (commented):
## Self-Review round 2 — clean

Reviewed immutable commit `cdd8d09af8aabee9320e632da2bcd44336658b31` from scratch across the complete base-to-head diff and its compiler-driver, typia/core/SDK composition, declaration/JavaScript/map/build-info/manifest publication, no-emit and failure recovery, concurrency, Windows/POSIX, Express/Fastify, generated SDK/Swagger/e2e, packaging, documentation, migration, and CI consequences.

No further sound improvement survived verification. The round confirmed:

- check, explicit `--noEmit`, and configured `noEmit` run the complete composed transform and publish no artifacts;
- LLM-incompatible `TypedRoute` return types report the exact `nestia.core.TypedRoute` compile diagnostic instead of deferring to runtime `NoTransformConfigurationError`;
- failed emitting builds publish no declarations, JavaScript, maps, build info, or manifest;
- the SDK contributor shares the core host and SDK callers preserve the host's nonzero status and stderr, so no separate SDK execution patch is required;
- buffered declaration callbacks are serialized before the final clean-only flush;
- typia, `@typia/interface`, and `@typia/utils` resolve consistently to 14.0.2.

GitHub Actions completed all 21 pull-request checks successfully, including `go`, `sdk`, `transform-options`, `e2e`, Ubuntu build, both format jobs, and both Ubuntu/Windows benchmark-system jobs. No verification remains skipped beyond the already disclosed interrupted local full-Go run, whose complete matrix was delegated to and passed in CI.

Issue #1627, opened 2026-08-25
Padding oracle in @EncryptedBody: decrypt failure (400) vs JSON parse failure (500) leaks PKCS#5 validity


samchon:
## Problem

`@EncryptedBody()` (`@nestia/core`) decrypts the request body with **AES-CBC + PKCS#5 padding** through `AesPkcs5` (`@nestia/fetcher`), which has **no authentication tag**. The decorator returns two *different* HTTP status codes for two different failure modes of a forged ciphertext:

- **decrypt failure** (bad PKCS#5 padding) is caught and turned into `400 Bad Request`;
- **`JSON.parse` failure** on a ciphertext that decrypts to valid-but-non-JSON bytes is **not** caught and propagates as `500 Internal Server Error`.

That status divergence is a classic **padding oracle**. An attacker who can submit modified ciphertexts (a MITM on the extra encryption layer, or any client able to POST to the endpoint) learns one bit per request — "was the PKCS#5 padding valid?" — and recovers plaintext **byte by byte without the key** (~256 requests per byte).

> Note on scope: this encryption layer is, by the maintainer's own statement in #253, an obfuscation layer meant to sit **on top of TLS** (block casual F12 inspection; satisfy a Korean financial-sector "encrypt once more" rule), not a replacement for TLS. The oracle is still a real defect: it removes the obfuscation guarantee entirely and returns a `500` for what is really malformed *client* input.

Reported privately as GHSA-pqj4-gvf7-6fq5 (@Fcmam5); filing the public tracking issue for the fix.

## Evidence (reproduced live)

A NestJS app using the real `@EncryptedController` + `@EncryptedBody`, key `"A".repeat(32)`, iv `"B".repeat(16)`:

```
ciphertext        : <48-byte AES-256-CBC ciphertext of a valid JSON body>
intact            -> 201   (decrypts, parses, validates)
flip last byte    -> 400   (PKCS#5 padding broken -> AesPkcs5.decrypt throws -> caught)
flip first byte   -> 500   (padding still valid -> JSON.parse throws -> uncaught)
256 last-byte guesses of the preceding block -> exactly 1 response != 400
```

The single non-`400` response in 256 guesses is the oracle: it pins the intermediate byte, and CBC gives the plaintext byte by XOR.

### Root cause

`packages/core/src/decorators/EncryptedBody.ts`:

```ts
const data: any = JSON.parse(decrypt(body, password.key, password.iv));
[...]

PR #1628, opened 2026-08-25
fix(core): close @EncryptedBody padding oracle

samchon:
Closes the padding-oracle vector reported in GHSA-pqj4-gvf7-6fq5 and tracked in #1627.

## Intent

`@EncryptedBody` returned `400` when AES-CBC/PKCS#5 decryption failed but `500` when the decrypted bytes were not valid JSON. That status divergence is a padding oracle: an attacker submitting forged ciphertexts recovers plaintext byte by byte without the key. This PR makes every malformed-input failure return one identical `400`, closing the oracle without touching the wire format.

## Scope (owns #1627)

- **Fix:** collapse decrypt failure and `JSON.parse` failure in `EncryptedBody` into a single identical `BadRequestException` (same status, message, and body).
- **Regression:** an e2e test in `tests/test-sdk/features/encrypted` that forges both failure modes and asserts the two responses are identical and neither is `500`.
- **Docs:** stop claiming "transport-level confidentiality" in `TypedBody.mdx` / `TypedRoute.mdx`; describe the obfuscation-over-TLS purpose and the unauthenticated-CBC residual.

## Deferred (breaking, not in this PR)

Unauthenticated CBC with a fixed IV (`IEncryptionPassword.iv`) is the deeper root cause. A real fix (AEAD + per-message random IV) changes the wire format and breaks every deployed `v<=13` peer, so it belongs to a maintainer-owned major release. Documented as residual on #1627 and in the guides.

## Non-breaking

Wire format unchanged; legitimate SDK traffic never reaches the failure path. The only observable change is forged/malformed input returning `400` instead of `500`.

## Verification

Pending — recorded as PR reviews as it lands.

### DIFF
--- a/packages/core/src/decorators/EncryptedBody.ts
+++ b/packages/core/src/decorators/EncryptedBody.ts
@@ -69,17 +69,35 @@ export function EncryptedBody<T>(
         : param;
 
     // PARSE AND VALIDATE DATA
-    const data: any = JSON.parse(decrypt(body, password.key, password.iv));
+    const data: any = decode(body, password.key, password.iv);
     const error: Error | null = checker(data);
     if (error !== null) throw error;
     return data;
   })();
 }
 
-/** @internal */
-const decrypt = (body: string, key: string, iv: string): string => {
+/**
+ * Decrypt and JSON-parse the request body behind a single failure response.
+ *
+ * `AesPkcs5` is AES-CBC without an authentication tag, so a MITM attacker can
+ * forge ciphertexts. Splitting the two failure modes of a forged body — a bad
+ * PKCS#5 padding (decryption throws) versus valid padding whose plaintext is
+ * not JSON (`JSON.parse` throws) — into two different HTTP status codes turns
+ * this decorator into a padding oracle: one observable bit per request lets the
+ * attacker recover the plaintext byte by byte without the key
+ * (GHSA-pqj4-gvf7-6fq5).
+ *
+ * Both failure modes must therefore collapse into one indistinguishable
+ * `BadRequestException` (identical status, message, and body). Only the
+ * downstream type validation of an already-parsed body may report its own
+ * distinct error, because reaching it requires plaintext the attacker cannot
+ * forge without already knowing it.
+ *
+ * @internal
+ */
+const decode = (body: string, key: string, iv: string): unknown => {
   try {
-    return AesPkcs5.decrypt(body, key, iv);
+    return JSON.parse(AesPkcs5.decrypt(body, key, iv));
   } catch (exp) {
     if (exp instanceof Error)
       throw new BadRequestException(
--- a/tests/test-sdk/features/encrypted/src/test/features/api/test_api_encrypted_padding_oracle.ts
+++ b/tests/test-sdk/features/encrypted/src/test/features/api/test_api_encrypted_padding_oracle.ts
@@ -1,0 +1,96 @@
+import { TestValidator } from "@nestia/e2e";
+import crypto from "crypto";
+
+import api from "@api";
+
+/**
+ * Verifies @EncryptedBody exposes no padding oracle: a forged ciphertext whose
+ * PKCS#5 padding is broken and one whose padding is valid but whose plaintext
+ * is not JSON must produce the identical failure response.
+ *
+ * `AesPkcs5` is unauthenticated AES-CBC, so a body can be forged. The decorator
+ * used to catch the decrypt failure as a 400 while letting the JSON.parse
+ * failure escape as a 500. Those two observable status codes are a padding
+ * oracle (GHSA-pqj4-gvf7-6fq5): one bit per request recovers plaintext byte by
+ * byte without the key. Both failure modes must now collapse into one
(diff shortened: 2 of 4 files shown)


samchon (commented):
## Commit `fix(core): close @EncryptedBody padding oracle` — ledger

Landed in one commit (owns #1627):

- **Fix** `packages/core/src/decorators/EncryptedBody.ts`: `decrypt()` (decrypt-only, caught → 400) + bare `JSON.parse` (uncaught → 500) replaced by a single `decode()` that runs decrypt **and** parse inside one try/catch, throwing one identical `BadRequestException`. Both forged-input failure modes are now indistinguishable (status + message + body).
- **Regression** `tests/test-sdk/features/encrypted/.../test_api_encrypted_padding_oracle.ts`: an independent-oracle test (encrypts with Node `crypto`, not the code under test), POSTs raw ciphertext to `/sellers/authenticate/login`, forges broken-padding and valid-padding-non-JSON bodies, and asserts both return an identical 400 and neither is 500.
- **Docs** `TypedBody.mdx` / `TypedRoute.mdx`: dropped the "transport-level confidentiality" claim; describe the obfuscation-over-TLS purpose and the unauthenticated-CBC limitation.

### Local verification (Node 24.18.0, Go 1.26.0, Windows)

- `pnpm --filter @nestia/core build` → exit 0.
- `pnpm --filter ./tests/test-sdk start -- --only encrypted` → both `encrypted` and `route-manual-validate-log-encrypted` pass; `test_api_encrypted_padding_oracle` runs (32 ms) and passes.
- **Mutation (sensitivity):** reverting only the `decode` unification (restore split decrypt / escaping `JSON.parse`) makes exactly `test_api_encrypted_padding_oracle` fail at "json failure is not a 500" (`x:400, y:500`) while all other encrypted tests still pass. Fix restored afterward.

### Deferred (not in this PR)

Unauthenticated CBC + fixed IV (`IEncryptionPassword.iv`) is the deeper root cause; a real fix (AEAD + per-message random IV) changes the wire format and breaks every deployed `v<=13` peer, so it belongs to a maintainer-owned major release. Recorded as residual on #1627.

CI pending on this head.

samchon (commented):
## Self-Review — round 1 (clean)

Full base-to-head diff re-read against the Non-Negotiable Review Law:

- **Correctness:** legitimate encrypted JSON is unchanged (`JSON.parse(AesPkcs5.decrypt(...))` runs the same); only the two forged-input failure exits are merged. Both throw the same `BadRequestException("Failed to decrypt ...")`, so status (400), NestJS body (`{statusCode,message,error}`), and message are byte-identical — no residual side channel via code, text, or length.
- **Consequence surface:** oracle exists only on the server request-decrypt path. `EncryptedRoute` encrypts responses (client decrypts, no server status oracle) and `EncryptedFetcher.decode` runs client-side — both correctly untouched. No generated SDK/Swagger baseline changes (encrypted bodies are absent from Swagger; the analyzers key on the decorator name, not this code).
- **Adapters:** Express (`raw-body`) and Fastify (`request.body`) both feed the one `decode()`; the fix is adapter-agnostic. The `text/plain` guard upstream is unchanged.
- **Test quality:** independent oracle (Node `crypto`, not the code under test); asserts the positive (valid ciphertext still 200/201) and the two negatives collapse to one identical 400; proven sensitive by mutation (reverting only the fix fails exactly this test at the 500 leak while all sibling tests pass).
- **Docs/voice:** `TypedBody`/`TypedRoute` no longer over-claim confidentiality; one-line-paragraph and plain-voice conventions kept; `spell-check` green.
- **Residual:** weak timing difference + fixed-IV determinism remain, both requiring the breaking AEAD change deferred to a major release (noted on #1627).

No sound improvement found. Integration gate: this is a decorator-runtime change; `origin/master` is unchanged since branch, and the PR runs the full suite (e2e/sdk/cli/editor/migrate/go/transform-options) on this exact head, which is the prospective squash-merge result. Will re-confirm `origin/master` is unmoved before merge and that all required checks are green.

Fcmam5:
@samchon , @claude PLEASE read my comment and email!

samchon (commented):
## Self-Review — round 2 (clean)

Adversarial re-audit of the whole diff, concurrent with CI:

- **Root-cause class swept (no whack-a-mole):** `EncryptedBody` is the only `@nestia/core` code that decrypts **and** `JSON.parse`s a request body (grep of `AesPkcs5.decrypt`/`JSON.parse` over `packages/core/src`). `EncryptedRoute` encrypts responses server-side (client decrypts — no status oracle); no sibling decorator shares the pattern.
- **No stray caller:** the old `decrypt` helper is fully replaced by `decode`; the only remaining reference is inside `decode` itself.
- **Validation-path is not a second oracle:** the distinct detailed 400 from `checker` is reachable only when the forged bytes are valid JSON, which requires plaintext the attacker cannot forge without already knowing it.
- **Boundary:** empty/short bodies still fail through `decode` to the same 400 (decrypt throws first); the `text/plain` guard is unchanged.

No sound improvement. Rounds 1–2 clean over head 2c184e33c.

PR #1637, opened 2026-09-16
ci(release): build every package before publishing and ship README

samchon:
## Intent

Every npm release since the monorepo (v12.0.0 through v13.0.3) shipped without a README, and the publish flow could leave npm half-released if a package build failed midway. Fix both by moving to a build-everything-first, publish-everything-after flow.

## Cause

- `packages/*/README.md` is gitignored; only local `pnpm release` ran `deploy/README.bash`, and the CI `publish` job never did, so each tarball's `files` entry for `README.md` silently matched nothing.
- `pnpm -r publish` triggered each package's `prepack: pnpm run build` right before publishing that package. A build failure in package N happened after packages 1..N-1 were already published.

## Change

- New root `package:prepare` = `pnpm build && node deploy/copy-readme.cjs`. `package:latest`, `package:next`, `package:rc`, and `package:tgz` all run it first, then publish or pack. A build failure now aborts before anything is published.
- `deploy/copy-readme.cjs` (Node, cross-platform) replaces `deploy/README.bash`; it copies the root README into every `packages/*` directory once.
- Removed `prepack: pnpm run build` from all eight packages so publish never rebuilds per package.
- `pnpm release` no longer copies README itself; the packaging step owns it.
- Updated `.agents/skills/{project,documentation}/SKILL.md` to describe the flow.

`release.yml` needs no change: its `publish` job calls `pnpm run package:${NPM_TAG}` and the website jobs call `pnpm run package:tgz`, both of which now include the build and README copy.

## Verification

- Confirmed the defect against the registry: `npm pack @nestia/core@13.0.3 --dry-run` lists LICENSE but no README.md (same for sdk, fetcher, migrate); `@nestia/core@4.6.0` did include it.
- Ran `pnpm package:tgz` on a clean state (all `packages/*/README.md` removed first): exit 0, README copied into all 8 packages, and all 8 tarballs in `deploy/tarballs/` contain `package/README.md`.

Already-published 13.0.3 stays as is; the next tag will be the first release with README included.


samchon (commented):
Follow-up after review: the first commit still let one lifecycle script run per package during publish.

**Gap.** pnpm runs `prepare` on publish/pack as well as `prepack`. `@nestia/migrate` has `prepare: pnpm run bundle`, which `git clone`s the template repositories from GitHub. So even with `prepack` gone, a network-bound script ran in the middle of `pnpm -r publish`, and a failure there would have left earlier packages already published.

**Fix (second commit).**
- `pnpm -r publish --ignore-scripts` for `package:latest|next|rc`.
- `deploy/tarballs` packs with `pnpm --config.ignore-scripts=true pack` (`pnpm pack` 10.6.4 rejects the `--ignore-scripts` flag form; the config form is honored).
- `package:prepare` now runs `pnpm --filter @nestia/migrate bundle` explicitly before `pnpm build`, so the bundles are regenerated after any version bump and compiled into `lib/` before packing. Previously the publish-time `prepare` re-stamped `src/bundles` only after `lib/` had already been built.

**Verification.**
- `pnpm publish --ignore-scripts --dry-run` in `packages/migrate`: no `prepare`/`bundle` output, README.md in the tarball listing.
- Full `pnpm package:tgz` from a state with no `packages/*/README.md`: exit 0; log order is bundle → build (8 packages) → copy-readme → 8× pack; zero `prepare`/`prepack`/`bundle` lines after the first `Building package (tgz)`; all 8 tarballs contain `package/README.md`.

samchon (commented):
Follow-up: moved the migrate bundle step out of the root script.

`pnpm --filter @nestia/migrate bundle` in root `package:prepare` was a package-internal build concern leaking into the workspace script. It now lives at the front of `@nestia/migrate`'s own `build` (`pnpm run bundle && rimraf lib && ttsc && ...`), and root `package:prepare` is back to `pnpm build && node deploy/copy-readme.cjs`. The ordering guarantee is unchanged: bundles are regenerated after any version bump and before `lib/` is compiled, and nothing runs during publish.

Verified `pnpm build` inside `packages/migrate`: bundle (clone nest + sdk templates) → ttsc → tsc declarations → rolldown → verify-package-exports, exit 0.

samchon (commented):
Follow-up: simplified the publish scripts.

- Dropped the `bash -c '... "$@"' --` wrapper. `pnpm run package:latest --no-git-checks --provenance` appends the extra arguments after the last command of the `&&` chain, so they land on `pnpm -r publish` without any wrapper. Verified under the pinned pnpm 10.6.4 with a scratch package, and the real command expands to `pnpm package:prepare && pnpm --filter=./packages/* -r publish --ignore-scripts --tag latest "--dry-run" "--no-git-checks"`. The wrapper came in with d49a6dae without a stated reason; before that the script was the plain form.
- Dropped `--access public`; all eight packages already declare `publishConfig.access: "public"`. A per-package `pnpm publish --dry-run` without the flag still lists the tarball normally.
- `pnpm run package:latest --dry-run --no-git-checks` on this branch: full build succeeds, README is copied, and pnpm then reports "There are no new packages that should be published" because 13.0.3 is already on the registry, which is the expected stop for a dry-run at the current version.

samchon (commented):
Follow-up: replaced `--ignore-scripts` with the structural fix.

The flag existed only because `@nestia/migrate` still had `prepare: pnpm run bundle`, and pnpm runs `prepare` on publish/pack as well as on install. Since the bundle step now runs at the front of migrate's `build`, that hook is redundant. Removed it, and removed `--ignore-scripts` from the three publish scripts and `--config.ignore-scripts=true` from `deploy/tarballs`.

Result: no `packages/*` declares any publish-time hook (`prepack`, `prepare`, `prepublishOnly`, `postpack`), so nothing can run during publish by construction. Verified `pnpm pack` in `packages/migrate` with no flags: only the file listing, no script lines.

Issue #1640, opened 2026-09-23
swagger schemas carry null for fields typia leaves absent, because the SDK bake marshals typia schemas with encoding/json


samchon:
## Problem

Generated Swagger documents contain `null` for schema fields that typia leaves absent. A literal union gets `"title": null` and `"description": null` on every `const` member, and an `any` property gets `"type": null`. Neither is valid: OpenAPI 3.1/3.2 (JSON Schema 2020-12) requires `title` and `description` to be strings and `type` to be a string or an array of strings. Strict validators, code generators and Swagger UI all read the document, and the CLI (`nestia swagger`) and the runtime `NestiaSwaggerComposer` produce it the same way.

Expected: the field is omitted, the way typia omits it in its own emitted literals.

## Evidence

Compile with `NESTIA_SDK_TRANSFORM=1` and generate Swagger for:

```ts
interface IQuery {
  literal: "x" | "y";
  anything: any;
}

@Controller("samples")
export class SampleController {
  @TypedRoute.Get()
  public async index(@TypedQuery() query: IQuery): Promise<void> {}
}
```

`components.schemas.IQuery.properties` comes out as:

```json
{
  "literal": {
    "oneOf": [
      { "const": "x", "description": null, "title": null },
      { "const": "y", "description": null, "title": null }
    ]
  },
  "anything": { "type": null }
}
```

The existing `exception` and `clone-*` features in `tests/test-sdk` show the same pattern (`"type": null`, and dozens of `"title": null` / `"description": null`) in their generated `swagger.json`.

Root cause: typia's Go JSON schema writer uses a Go `nil` map value to mean "absent" (`json_schema_constant.go` stores `json_schema_title(...)`, a `*string` that is nil when there is no `@title`; `json_schema_station.go` sets `output["type"] = nil` for `any`). Its own `LiteralFactory` documents and follows that contract ("Object writers treat a Go `nil` value as absent and skip it"; a real `null` uses `LiteralFactory_Null`). nestia's SDK contributor does not follow it. `nestiaSDKTryBakeJsonSchema` in `packages/sdk/native/sdk/sdk_transform.go` copies typia's `JsonSchema` maps into the metadata literal, and `nestiaSDKMetadataLiteralText` serializes them with `encoding/json`, which writes every nil as `null`. That includes nils nested inside `LiteralFactory_OrderedObject` values, whose own `MarshalJSON` only skips the object's direct nil entries.

[...]

Issue #1639, opened 2026-09-22
Decomposed query/header parameters drop every typia tag in the generated swagger


bartekburaczewski:
### Summary

Since `@nestia/sdk` v13, typia tags on `@TypedQuery` / `@TypedHeaders` properties disappear from the generated OpenAPI document. `format`, `minimum`, `maximum`, `default` and `minItems` are all lost. Request body, response and `@TypedParam` schemas are unaffected.

- **SDK version**: reproduced on 13.0.2 and 13.0.4, and the cause is still on `master`
- **Other versions**: `typia` 14.0.4, TypeScript 7.0.2, `@nestjs/*` 12.0.4
- **Expected behavior**: decomposed query parameters carry the JSON schema constraints their typia tags declare
- **Actual behavior**: every parameter is reduced to its bare atomic type

I hit this on a real backend upgrading from `@nestia/sdk` 11.3.4 + `typia` 12.1.1, where 19 query parameters lost `format`, 16 lost `default` and `minimum`, 8 lost `maximum` and one array parameter lost `minItems`.

### Code occurring the bug

```typescript
import { TypedQuery, TypedRoute } from "@nestia/core";
import { Controller } from "@nestjs/common";
import type { tags } from "typia";

interface IQuery {
  from: string & tags.Format<"date-time">;
  limit: number & tags.Minimum<1> & tags.Maximum<100> & tags.Default<10>;
}

@Controller("samples")
export class SampleController {
  @TypedRoute.Get()
  public async index(@TypedQuery() query: IQuery): Promise<void> {
    void query;
  }
}
```

Built with `NESTIA_SDK_TRANSFORM=1 ttsc`, then `NestiaSwaggerComposer.document(app, { openapi: "3.0" })`. The parameters of `GET /samples` come out as:

```json
[
  { "name": "from", "in": "query", "schema": { "type": "string" }, "required": true },
  { "name": "limit", "in": "query", "schema": { "type": "number" }, "required": true }
]
```

Expected:

```json
[
  { "name": "from", "in": "query", "schema": { "type": "string", "format": "date-time" }, "required": true },
  { "name": "limit", "in": "query", "schema": { "type": "number", "minimum": 1, "maximum": 100, "default": 10 }, "required": true }
]
```

### Cause

[...]

Issue #1643, opened 2026-09-23
the TypedQuery guide says swagger.decompose is opt-in and shows a nested query DTO that does not compile


samchon:
## Problem

The website guides describe `swagger.decompose` wrongly in three places:

1. **The default is reversed.** `website/src/content/docs/core/TypedQuery.mdx` ("Nested DTOs and Swagger decomposition") says "By default, Swagger documents the whole object as a single `query` schema. To **decompose** it ... set `swagger.decompose: true`". The default has been `true` since dee400980 ("Close #943: `config.swagger.decompose` to be default `true`", 2024-07). `INestiaConfig.ISwaggerConfig.decompose` documents `@default true`, and `SwaggerOperationParameterComposer` decomposes unless `decompose === false`.
2. **The example doesn't compile.** The same section says "nested objects work too" and shows a query DTO with a nested `filter?: { section?: string; after?: string & tags.Format<"date-time"> }` object. `@TypedQuery()` rejects it at compile time:

   ```text
   error TS(nestia.core.TypedQuery): unsupported type detected
   - ISearchQuery.filter: (__type | undefined)
     - nested object type is not allowed.
   ```

   typia's HttpQuery validator allows only atomic, constant, template-literal and array-of-those property types.
3. **The configuration reference is vague.** `website/src/content/docs/swagger/index.mdx` describes `decompose` as "Split query-DTO into one parameter per leaf field." It doesn't state the default. It doesn't say that header objects are decomposed too. "Leaf field" suggests a nesting that can't exist. The page's "See also" also sends readers to **Swagger → Strategy** for "decomposition", but `strategy.mdx` never mentions it.

## Evidence

- `grep -n decompose packages/sdk/src/INestiaConfig.ts` shows `@default true`.
- `git show dee400980` flips the default.
- Compiling the guide's `ISearchQuery` with `@TypedQuery()` gives the diagnostic above.
- `SwaggerOperationParameterComposer.decomposible` handles `category === "query"` and `"headers"` alike.

## Consequence surface

Readers who follow the guide either think they must opt in to a behavior they already have, or copy a DTO that fails to build. Readers of the configuration table can't tell what the default is or that headers are affected.

## Approach

The guides should describe what the generator does: `decompose` defaults to `true` and applies to query and header objects, query DTO properties must be flat, and decomposed parameters carry the property's constraints. The guide changes should land with the generator fixes they describe (#1639 and the related decomposition issues).

## Acceptance and verification

[...]

Issue #1642, opened 2026-09-23
decomposed query/header parameters drop @deprecated and @SwaggerExample.Parameter() examples


samchon:
## Problem

When `swagger.decompose` splits a `@TypedQuery()` / `@TypedHeaders()` (or plain `@Query()` / `@Headers()`) object into one parameter per property, the document loses two things the undecomposed form keeps:

1. **`@deprecated` properties.** typia marks a deprecated property with `deprecated: true` in the object schema. The decomposed parameter carries no `deprecated`, although the OpenAPI Parameter Object defines that field. nestia already maps route-level `@deprecated` to `Operation.deprecated`.
2. **`@SwaggerExample.Parameter()` examples.** With `decompose: false`, the query or header parameter carries the decorator's `example` / `examples`. With the default `decompose: true`, they are silently dropped. The SwaggerExample guide says the decorator stacks with `@TypedQuery()`, and `decompose` defaults to `true`, so the documented usage produces no example at all.

## Evidence

```ts
export interface ISmallQuery {
  keyword: string & tags.MinLength<1>;
  /** @deprecated */
  page?: number & tags.Type<"uint32">;
}

@Controller("example")
export class ExampleController {
  @TypedRoute.Get("query")
  public async exampleQuery(
    @SwaggerExample.Parameter({ keyword: "nestia", page: 3 })
    @SwaggerExample.Parameter("named", { keyword: "typia" })
    @TypedQuery()
    input: ISmallQuery,
  ): Promise<void> {}
}
```

The generated parameters are:

```json
[
  { "name": "keyword", "in": "query", "schema": { "type": "string" }, "required": true },
  { "name": "page", "in": "query", "schema": { "type": "number" }, "required": false }
]
```

There is no `deprecated` on `page` and no `example` / `examples` on either parameter. The lost schema constraints are #1639.

Root cause: the decomposition map in `SwaggerOperationParameterComposer.decomposible` (`packages/sdk/src/generates/internal/SwaggerOperationParameterComposer.ts`) builds each parameter from `name`, `in`, `schema`, `required` and `description` only. The undecomposed branch of the same function passes `example: props.parameter.example, examples: props.parameter.examples`, but the decomposed branch never reads them or the property's `@deprecated` tag.

## Consequence surface

| Input | Undecomposed today | Decomposed today | Decomposed expected |
| --- | --- | --- | --- |
| property `@deprecated` | property schema `deprecated: true` | nothing | parameter `deprecated: true` |
[...]

PR #1644, opened 2026-09-23
fix(sdk): describe query and header parameters exactly as typia does

samchon:
## Intent

This PR owns the whole accepted cycle of the issue campaign started from #1639. The goal is that Swagger's query and header parameters describe the API exactly as typia describes the underlying types, with nothing dropped by decomposition.

| Issue | Defect | Owner |
| --- | --- | --- |
| #1639 | decomposed query/header property schemas come from a lossy JS fallback: tags, integer types, template patterns and constant docs are lost, literal objects get a dangling `$ref`, and `@internal` / function-only properties leak in | SDK Go bake + `SwaggerOperationParameterComposer` |
| #1640 | the SDK bake serializes typia's nil-as-absent fields as `null` (`title`, `description`, `type`) | SDK Go bake |
| #1641 | field-named `@Headers("x")` parameters are missing from operations | `SwaggerOperationComposer` |
| #1642 | decomposed parameters drop `@deprecated` and `@SwaggerExample.Parameter()` examples | `SwaggerOperationParameterComposer` |
| #1643 | the guides misstate `decompose` (default, header coverage, a nested DTO that doesn't compile) | website guides |

## Status

Verification is pending. This is a claim commit; implementation, tests, the consequence matrix and verification records will follow as formal review comments.

## Out of scope (observed, not changed here)

- The SDK generator shadows a function's namespace when a controller method and its parameter share a name (`query(query)` → `query.METADATA` resolves to the parameter).
- `legacy.ts` `HttpQueryProgrammer` / `HttpHeadersProgrammer` validators are no-ops, so plain `@Query()` / `@Headers()` DTOs with nested objects pass SDK analysis, where v11 rejected them.
- `SwaggerExample` named `examples` are emitted as raw values rather than OpenAPI Example Objects, across bodies, parameters and responses.


samchon (commented):
**Commit 3b4035a70: native bake**

- The resolved schema of every route parameter now carries `jsonSchema.properties`: typia's value schema of each property of its first object type. It is written by typia's own `Json_schema_station` against the parent's components. Properties typia's object schema omits (`@hidden`, `@ignore`, `@internal`, or a value with no JSON form) get no entry. Primitive schemas, responses and exceptions carry none.
- Baked schemas no longer print typia's absent nils as `null`. A nil member of a schema object is dropped. Instance keywords (`const`, `default`, `enum`, `example`, `examples`) and `x-` extensions keep a nil as `null`, because the pinned typia (`f692fd63a846`, typia 14.0.0) writes a declared null there as a bare nil too. `JsonSchemaPlugin<{ "x-empty": null }>` and `Examples<{ none: null }>` are the regression twins.
- Go regressions: `TestSyntheticParameterPropertySchemasBakeTypiaValueSchemas` and `TestSyntheticJsonSchemaOmitsTypiaAbsentFields`.

Resolves #1640. #1639's native half landed; its JS consumer follows in the next commit.

samchon (commented):
**Commit 9d88906da: Swagger composition**

- `SwaggerOperationParameterComposer` builds decomposed query and header parameters from `jsonSchema.properties` through the new `JsonSchemasProgrammer.writeProperty`, instead of the atomic-only fallback. A property without a baked entry, which typia's object schema omits, yields no parameter, and `@internal` now joins `@hidden` and `@ignore`. The fallback remains only for metadata the transform did not bake.
- A decomposed parameter carries `deprecated` for a `@deprecated` member. It also takes its own member of the object's `@SwaggerExample.Parameter()` default and named examples, in the representation undecomposed parameters already use.
- `SwaggerOperationComposer` lists `headerParameters` (field-named `@Headers("x")`) again, between the query object and the header object.
- New feature `tests/test-sdk/features/swagger-parameters`, one assertion per file:
  - `test_swagger_decomposed_query_schemas` and `test_swagger_decomposed_headers_schemas`: typed and plain decorators, oracle = typia's component schema minus property-level fields, plus pinned tag cells.
  - `test_swagger_decomposed_property_set`: names and required flags equal the component's.
  - `test_swagger_decomposed_deprecated` and `test_swagger_decomposed_examples`.
  - `test_swagger_field_headers`.
  - `test_swagger_absent_schema_fields`: the only nulls in the whole document are the declared `examples` / `x-` values.
  - `test_swagger_decomposed_downgrade`: 3.1, 3.0 and 2.0.
- Local: `TEST_SDK_SKIP_BUILD=1 pnpm --filter ./tests/test-sdk start -- --only swagger-parameters` passes.

Resolves #1639, #1641 and #1642.

samchon (commented):
**Commit 6a740a68b: dynamic-key objects (#1645)**

The implementation pass found that a plain `@Headers()` / `@Query()` object with a dynamic key crashes Swagger generation (`Cannot read properties of undefined (reading 'values')`). That is the `Record<string, string>` pattern `core/TypedHeaders.mdx` recommends. Filed as #1645.

- An object decomposes now only when every key is a sole literal (`isSoleLiteralOf`). Otherwise the undecomposed parameter, carrying typia's object schema with `additionalProperties`, is emitted. Splitting off the known keys would drop the dynamic part.
- Regression: `test_swagger_dynamic_key_objects` covers pure-`Record` headers, pure-`Record` query, and a mixed static-plus-index-signature query. The document only exists if generation did not crash.
- The fixture handlers are not named `query` / `headers`, to avoid the out-of-scope SDK shadowing noted in the PR body.

Resolves #1645.

samchon (commented):
**Commit 533245e42: early-warning follow-ups**

Adjudication of the early-warning findings on 3b4035a70 and 9d88906da. Each was reproduced before any action.

| Finding | Disposition |
| --- | --- |
| Property `x-` JSDoc extensions missing from the bake, which `decompose: false` keeps | **Accepted.** The bake applies typia's `json_schema_jsDocTags` rule (unexported, replicated as `nestiaSDKJsDocExtensions`). The decomposed schema is now typia's property schema minus title / description / deprecated / readOnly. This re-adjudicates my earlier rejection, and the test oracle states the invariant exactly. |
| Instance-keyword nil rule untested | **Accepted.** `JsonSchemaPlugin<{ example: null; default: null }>` added to the Go test. |
| Recursion into `items` / `prefixItems` / `additionalProperties` untested | **Accepted.** Array, tuple and record literal unions added to the Go test. |
| Swagger 2.0 gets a parameter `deprecated` | **Accepted.** The flag is omitted when `swagger.openapi` is `"2.0"`, following the server-description precedent in `SwaggerGenerator.initialize`. Pinned by `openapi_v2/test_openapi_v2_decomposed_parameters`. |
| Baked schemas shared by reference across compositions | **Accepted.** `writeProperty` returns a copy, restoring the isolation decomposed schemas had. |
| `@internal` JS filter / fallback untested; `@hidden` fixture actually `@ignore` | **Accepted.** `prettier-plugin-jsdoc` rewrites `@hidden` to `@ignore` in sources. `test_swagger_decomposed_fallback` registers hand-written metadata with and without the bake. `@hidden` is also pinned in the Go test. |
| A header listed both as a field and inside a header object is emitted twice | **Out of scope, reported.** The same duplicate already exists for field queries plus a query object. The contradiction belongs to SDK route analysis, not to decomposition. |
| Named `examples` are raw values, not Example Objects | **Out of scope, reported.** Undecomposed parameters, bodies and responses already emit raw values, and `@nestia/migrate` reads them back raw into `SwaggerExample.*(key, value)`. Changing the shape is a cross-package contract change. Decomposed parameters keep the existing representation. |

Local: `pnpm --filter @nestia/sdk test:go` (both new tests), and `swagger-parameters` / `openapi_v2` via `TEST_SDK_SKIP_BUILD=1 pnpm --filter ./tests/test-sdk start -- --only <feature>`, all pass.

samchon (commented):
**Commit c00c49e78: guide follow-ups (#1643)**

From the early-warning pass on 07cd1520a:
- `core/TypedHeaders.mdx` no longer lists SDK visibility as an advantage over `@Headers()`. The closing note says both reach the SDK and Swagger, and that a dynamic-key header object contributes only its known header names, matching 0785375ca.
- `core/TypedQuery.mdx` states which members decomposition skips and that Swagger 2.0 cannot carry `@deprecated`.
- **Out of scope, reported:** `core/TypedBody.mdx`, `core/TypedParam.mdx` and `core/TypedRoute.mdx` make the same "vanilla decorator is invisible to `@nestia/sdk`" claim about bodies, path params and routes. Those pages are outside this cycle's query/header surface.

samchon (commented):
**Commit 0785375ca: dynamic keys, per format (#1645)**

The early-warning pass on 6a740a68b found the first fix right only for an OpenAPI 3.x query. A `Record<string, string>` header object became one required header literally named `headers`. Under `openapi: "2.0"` a dynamic query object downgraded to a `$ref` parameter, which 2.0 reads as a parameter reference.

- 3.x query objects with a dynamic key stay one form-style object parameter, the idiom that explodes into arbitrary keys.
- Header objects, and 2.0 query objects, contribute their statically keyed properties only.
- New and updated regressions:
  - `test_swagger_dynamic_key_objects`: pure and mixed headers, pure and mixed 3.x query.
  - `openapi_v2/test_openapi_v2_dynamic_query`: 2.0 mixed and pure query.
  - `test_swagger_decomposed_isolation`: pins the copy in `writeProperty`, which no test covered.
- #1645's body is revised to this acceptance.

Adjudication of the pass on 533245e42:
- Schema-level `example` / `title` spilling onto 2.0 parameters comes from typia's `SwaggerV2Downgrader`, which spreads the downgraded schema into the parameter. It already happens for path and field parameters on master. **Upstream**; nestia only suppresses parameter fields it adds itself (`deprecated`).
- `components` returned by reference matches master's `writeSchemas` path. **Not a regression.**
- `@x-foo NaN` failing the metadata marshal also fails the component on master. **Pre-existing**, out of scope.

samchon (commented):
**Commit ee2a79804: object `required` + Self-Review round 1 corrections**

- Early-warning on 0785375ca found no defects. It flagged that the 3.x object parameter for a `Record` query was `required: true` although the endpoint accepts no query string. **Accepted.** An undecomposed object parameter, whether `decompose: false` or the dynamic-key query object, is now required only when one of its named properties is. Twins: `/record/required` (required keyword → true), `/record/query` and `/record/mixed` (→ false), and `query-decompose-false/test_swagger_optional_object` (all-optional DTO → false, `IQuery` → true).
- The same pass flagged that `decompose: false` renders a header object as one header named after the handler parameter. **By design**, per #822: `decompose` is the documented remedy for that representation.
- **Self-Review round 1** (full base-to-head diff at c00c49e78) found no behavioral defect and three stale comments, corrected here:
  - `nestiaSDKTryBakeJsonSchema` said "value schema", but the bake carries the property schema including `x-` extensions.
  - The normalizer cited `tags.Example<null>` as its real-null example, but the pinned typia never attaches that tag.
  - The composer comment said "value schema".

samchon (commented):
**Commit 10a708c6d: guides.** From the early-warning pass on c00c49e78. `TypedHeaders.mdx` now says a decomposed header object's dynamic key adds no parameter (none for a pure `Record`, the named ones for a mixed object). `TypedQuery.mdx` says what 3.x and 2.0 do with a vanilla `@Query()` object that has a dynamic key, and that decomposition skips every property typia's JSON schema leaves out.

samchon (commented):
**Commits f6f730812 and 7fda92cbe: early-warning follow-ups (ee2a79804, 10a708c6d)**

- ee2a79804's pass raised three cases, all **accepted** and twinned in `query-decompose-false/test_swagger_optional_object`:
  - A field-named object (`@Query("filter")`) keeps its declared `required`.
  - A union of objects is optional if any member needs none of its keys.
  - Only described properties count; an `@internal` required member publishes nothing. The describability test is now the shared `isDescribed` predicate, the same one decomposition filters with.
- The same pass noted that a required `string[]` member keeps the object required although typia parses a missing key as `[]`. **Not a regression**: the decomposed form says the same, following typia's own schema.
- 10a708c6d's pass: the 2.0 dynamic-query note now says it holds when decomposing, and TypedHeaders links the decomposition section. **Accepted.**
- Self-Review round 2 (full diff at 10a708c6d) found one comment wording issue, `required` described as "an empty query string" while the rule covers headers too. Fixed in f6f730812.

samchon (commented):
**Self-Review round 3** at 7fda92cbe, full base-to-head diff: no finding. Early-warning on f6f730812 found no defect introduced. Its one candidate is a field-named object query (`@Query("filter") filter: IFilter`), emitted without `style: deepObject` and split by decomposition. That is master behavior, on an input the SDK's v11 validators rejected, so it joins the out-of-scope no-op-validator item in the PR body. Pending: mutation-sensitivity evidence, CI on this head, and the integration gate.

samchon (commented):
**Commit 4935953ef: mutation evidence and a corrected premise**

Mutation sensitivity was checked in a disposable worktree. Each product change was mutated alone, with the tests kept.

- **test-sdk (c00c49e78), 11/11 killed:** decompose ignores the bake, field headers omitted, no `deprecated`, 2.0 `deprecated`, no examples, no dynamic-key handling, no `@internal` filter, no copy, 2.0 dynamic query object. The control passes.
- **Go (c00c49e78), 8/8 killed:** normalizer identity, no property bakes, no `x-` JSDoc, no instance keywords, no `items` recursion, no `prefixItems` recursion, no hidden/ignore/internal skip, properties on every pipe. The control passes. A first Go run was vacuous because `cmd.exe` split the `-run` regex on `|`. It was re-run without a shell, and each run requires the named tests to actually execute.
- **`required` rule (7fda92cbe):** R1 declared-only, R2 field not exempt and R3 first union member were killed. R4 (count every literal key, not only described ones) **survived**.

R4 survived because typia's metadata factory drops `@internal` members before the SDK sees them, so an `@internal` twin cannot distinguish the rule. This commit switches the twin to an `@ignore` member (in metadata, out of schema). The same finding shows that my earlier "decomposition emitted `@internal` members" premise for #1639 was wrong for transform-produced metadata. It is corrected on #1639 and in the two affected test docblocks. The JS `@internal` filter stays, consistent with typia's object schema, but it is only reachable with hand-written metadata.

samchon (commented):
**Self-Review round 4** at 4935953ef, full base-to-head diff: **no finding.**

- Source and guides are unchanged since round 3. The only delta is the `@ignore` twin in `query-decompose-false` and the corrected test docblocks.
- Test oracles still derive from typia's own component schema, not from the composer under test. No assertion is vacuous: the per-parameter loops iterate names that `test_swagger_decomposed_property_set` pins.
- Mutation sensitivity is complete. test-sdk 11/11, Go 8/8 and the `required` rule 4/4 are all killed, with passing controls.
- CI: all 21 checks pass on this head.

Pending before merge: the local integration gate on this exact head (disposable worktree: `pnpm build`, `pnpm test`) and the clean packed-consumer smoke.

samchon (commented):
**Self-Review round 5** (fresh full read at 4935953ef, all 51 files) → fixed in this commit

Round 4 leaned on "source unchanged since round 3" instead of a full re-read, which the review law does not accept, so round 5 re-read everything: Go bake and normalizer, JS composers and `legacy.ts`, Go tests, test-sdk fixtures and tests, and guides.

Findings, both test and comment accuracy, no behavior:
1. Since 533245e42 a decomposed schema is typia's property schema minus the parameter's own fields, `x-` JSDoc extensions included. The oracle helper, two test docblocks, a DTO comment, and a Go test step and name still said "value schema". Renamed: `parameterSchemas`, `TestSyntheticParameterPropertySchemasBakeTypiaSchemas`.
2. The synthetic Go tests put nested-object, function, tuple and record members in `@TypedQuery()` DTOs, which core's HttpQuery validation rejects in a real build. The docblocks now say only the SDK contributor runs in-process and that the bake treats every parameter alike. They also note that typia's metadata drops `@internal` before the bake.

Re-checked with nothing found: the `Properties` cache key, the TypeGuardError early return, `Json_schema_station` returning fresh maps (so `x-` writes cannot touch components), the normalizer's keyword classes (a declared `@x-foo null` stays `null` in both the component and the parameter), the `writeProperty` fallback and idempotent component merge, and the fixtures (no handler shares its parameter's name).

A fresh round 6 follows at this head.

samchon (commented):
**Self-Review round 6** at 2607cd078, a fresh full read of all 51 files: **no finding.**

- Go: the bake, cache key and normalizer were re-read. Readonly-array marking of decomposed schemas comes from the JS emender on the copied schema, and `values` / `generic` in `test_swagger_decomposed_query_schemas` compare it against the component. Body parameters also get a property bake nothing reads, which is bounded to `NESTIA_SDK_TRANSFORM` output.
- JS: the `required` rule covers optional, nullable, dynamic-key and `decompose: false` header objects as intended. The decomposition path, examples and 2.0 handling are unchanged.
- Tests: each of the 15 test files' headline, rationale and assertions match the current behavior, and the oracles come from typia's component.
- Guides: all claims match the composer.

Gate note: the local root `pnpm test` stalled twice in `packages/core test:go` with a warm Go test cache. It reproduced with `go test ./...` (no output after 180 s) against `-count=1` (41 s). The cause is Go's test-result cache validation in `packages/core/test`, which this PR does not touch; CI runners have no cache. It is reported as out of scope, and the gate re-runs at this head with `GOFLAGS=-count=1`.

samchon (commented):
**Merge gate** at 2607cd078. `origin/master` is still 1d6ddd6c3, so the prospective merge result is this head.

| Gate | Result |
| --- | --- |
| CI | 21/21 checks pass |
| Self-Review | round 6 clean (fresh full read) |
| Integration gate (disposable worktree, Windows): `pnpm install --frozen-lockfile`, then `GOFLAGS=-count=1 pnpm test` (root build, core + sdk Go tests, all seven `tests/test-*` workspaces) | exit 0; core Go ok 42.3 s, sdk Go ok 35.2 s, test-sdk including `swagger-parameters`, `openapi_v2`, `query-decompose-false` |
| Mutation sensitivity | test-sdk 11/11, Go 8/8, `required` rule 4/4 killed; controls pass |
| Clean packed consumer: `pnpm package:tgz`, then a fresh npm project installing the tarballs per the setup guide and running `npx nestia swagger` | passed. The #1639 repro yields `format`, `minimum`/`maximum`/`default`, `integer` and `deprecated`; the field header `x-trace` is listed; no `null` anywhere; a `Record` header object adds no parameter |

Closing reconciliation: the commit messages close #1639, #1640, #1641, #1642, #1643 and #1645. There are no reverts in the range, every fix is present at HEAD, and the PR body carries no closing keyword.

`GOFLAGS=-count=1` was needed locally because a warm Go test cache stalls `packages/core/test` (reported as out of scope). It does not affect CI.

Issue #1658, opened 2026-09-23
Handover: resolve #1646–#1656 in one campaign pull request


samchon:
## Mandate

Resolve #1646 through #1656 as one issue campaign cycle in a **single pull request**, and merge it. Each linked issue is self-contained: repro, root cause, consequence surface, approach, and acceptance.

- **Start at implementation.** These issues are already adjudicated and published. Follow `.agents/skills/issue-campaign/development.md` together with the project, development, review, and pull-request skills.
- **Leave nothing behind.** Fix every defect you find in these issues' consequence surface in the same pull request. Listing it in a report does not count.
- **Open decisions are yours** (#1648, #1649, #1650, #1652). Decide by the issue's invariant and the existing precedent, and record the evidence in the pull request.
- **Before changing behavior, find out whether the current code is deliberate.** Check the types, the tests that pin it, commit and PR text, and doc comments, then state in the pull request why that reasoning does not cover the case. Several issues already cite such reasons: #1470's validator stubs, raw examples round-tripping through migrate, and the SDK escaping its own identifiers.
- **Out of scope:** releases (#1657) and every `version` field.
- **Finish.**
  - Run Self-Review rounds while CI runs.
  - Squash-merge only after green CI and a clean round.
  - The pull request closes #1646 through #1656 and this issue.
  - Report to the user in the user's language.

## Order

| Order | Issue | What | Where |
| --- | --- | --- | --- |
| 1 | #1646 | Go module pins out of sync with the npm dependency (verified diff in the issue) | `go.mod` / `go.sum`, Go |
| 2 | #1656 | `-count=1` for the three `go test` scripts | `package.json` |
| 3 | #1655 | NaN / ±Inf `@x-*` values abort generation | Go SDK bake |
| 4 | #1654 | `SwaggerCustomizer` edits leak across compositions | `legacy.ts` |
| 5 | #1647 | generated SDK does not compile on name collisions | SDK generators |
| 6 | #1650 | same key in a field and an object parameter | analyzer, Swagger, SDK |
| 7 | #1652 | `@UploadedFile()` / `@UploadedFiles()` / `@RawBody()` silently dropped | analyzer |
| 8 | #1648 | plain `@Query()` / `@Headers()` / `@Param()` unvalidated | analyzer / Go bake |
| 9 | #1649 | named examples are raw values; 2.0 generation fails | Swagger composers, migrate |
[...]

Issue #1656, opened 2026-09-23
test:go takes minutes on Windows: add -count=1 to go test


samchon:
## Problem

Running the Go tests locally on Windows takes several minutes and looks hung, while CI finishes the same package in seconds:

| Command | Time |
| --- | --- |
| `pnpm --filter @nestia/sdk test:go` (`cd test && go test ./...`), `TestTransformProjectModeEmitsAllFiles` alone | 178.08 s |
| same test with `go test -count=1` | 1.28 s |
| whole `packages/sdk/test` package, cached mode | > 300 s (interrupted) |
| whole package, `-count=1` | 41 s |
| CI (`ubuntu-latest`) | ~16 s |

## Root cause

`go test ./...` without `-count=1` runs in cacheable mode. To decide whether a result may be cached later, Go instruments the test binary (`internal/testlog`) and records every file the test opens and every environment variable it reads. The transform tests open thousands of files (the TypeScript program, `node_modules` typings). On Windows that bookkeeping is extremely slow, about 140× here. `-count=1` disables caching and the instrumentation with it. Caching is also wrong for these tests in principle: they read the pnpm-installed toolchain and fixtures outside Go's view, so a cached pass can be stale.

## Approach

Add `-count=1` to all three `go test` invocations:

- `packages/sdk/package.json` `test:go`: `cd test && go test ./...`
- `packages/core/package.json` `test:go`: `cd test && go test ./...`
- root `package.json` `test:go`: `... && cd packages/core/native && go test ./...`

CI runs the root `pnpm test:go` (`.github/workflows/test.yml`), so fixing the three scripts keeps CI and local runs identical.

Trade-off: on Linux and macOS, `-count=1` also gives up the cache that makes a repeated local run instant. That is acceptable, because these tests read toolchain and fixture files outside Go's view, so a cached pass can be stale.

## Acceptance and verification

- `pnpm --filter @nestia/sdk test:go` finishes in under a minute on Windows (was > 5 minutes).
- `grep -rn "go test" --include=package.json` shows `-count=1` everywhere.
- CI stays green.

Found during the #1639 campaign (#1644).

Issue #1655, opened 2026-09-23
`@x-*` tag value NaN / Infinity aborts SDK and Swagger generation (json: unsupported value)


samchon:
## Problem

A JSDoc `@x-*` extension tag whose value typia reads as a non-finite number makes SDK and Swagger generation fail:

```ts
export interface IItem {
  /** @x-level NaN */
  a: string;
  /** @x-level Infinity */
  b: string;
}
```

`nestia sdk` / `nestia swagger` (and any build with `NESTIA_SDK_TRANSFORM=1`) stops with

```
error TS(nestia.sdk.OperationMetadata): json: unsupported value: NaN
error TS(nestia.sdk.OperationMetadata): json: unsupported value: +Inf
```

`nan`, `inf`, `infinity`, and `-Infinity` fail the same way (`json: unsupported value: NaN` / `+Inf` / `-Inf`). One tag value anywhere in a controller's reachable types breaks generation of the whole project.

## Root cause

typia casts `@x-*` tag text to a number when it parses as one, and Go's `strconv.ParseFloat` accepts `NaN`, `Infinity`, `inf`, and so on. The SDK contributor (`packages/sdk/native/sdk/sdk_transform.go`) serializes the baked schema with `encoding/json`. `encoding/json` refuses NaN and ±Inf, and the error aborts the transform.

nestia must not crash on whatever values typia produces.

## Approach

Invariant: no JSDoc tag value can make generation fail with a serializer error. Encode non-finite numbers the way JavaScript's `JSON.stringify` does, as `null`. Do it in the SDK contributor's schema literal normalization (`nestiaSDKJsonSchemaLiteral` path), so every baked value goes through one rule. Apply it to every place the contributor serializes typia output, not only `x-` extensions.

## Acceptance and verification

- The interface above generates. Every non-finite value typia hands over is written as `null`, matching `JSON.stringify`.
- A Go test in `packages/sdk/test` covers `NaN`, `Infinity`, `-Infinity`, and `inf` tag values and asserts the baked literal.
- `pnpm test` is green.

Found during the #1639 campaign (#1644).

Issue #1654, opened 2026-09-23
SwaggerCustomizer edits accumulate across NestiaSwaggerComposer.document() calls


samchon:
## Problem

`SwaggerCustomizer` edits leak between Swagger documents composed in the same process. Every `NestiaSwaggerComposer.document()` call re-applies the customizers on top of the previous call's edits:

```ts
@SwaggerCustomizer((props) => {
  const schema = props.swagger.components.schemas!["IItem"] as OpenApi.IJsonSchema.IObject;
  schema.description = (schema.description ?? "Item.") + " Customized.";
})
@TypedRoute.Get()
public get(): IItem { … }
```

```ts
const app = await NestFactory.create(AppModule);
for (let i = 0; i < 3; ++i) {
  const doc = await NestiaSwaggerComposer.document(app, {});
  console.log(doc.components.schemas.IItem.description);
}
// Item. Customized.
// Item. Customized. Customized.
// Item. Customized. Customized. Customized.
```

Real users compose more than once per process: one document per version or per module, `SwaggerModule.setup` plus a file export, or tests that build the app repeatedly in one worker. Every repeated call corrupts the document, and any customizer that is not idempotent (appending, pushing to arrays, counters) gets worse each time.

## Root cause

The schemas baked into `OperationMetadata` by the SDK transform are module-level constants, shared by every composition. In `packages/sdk/src/internal/legacy.ts`:
- `JsonSchemasProgrammer.writeSchemas` pushes `baked.schema` by reference.
- It also `Object.assign`s the baked component schema *objects* into a fresh map, so the map is new but every schema in it is shared.
- #1644 made `writeProperty` deep-copy the property schema for this very reason, but `writeProperty` still returns the baked `components` object itself.

The customizer mutates the document, and with it the shared baked objects, so the next composition starts from the mutated objects.

## Consequence surface

- `NestiaSwaggerComposer.document()` at runtime (reproduced above);
- `nestia swagger` CLI generation: check whether any configuration composes more than one document per process (for example an array config), and cover it if so;
- any other consumer that mutates a composed document (for example post-processing in user code) — same shared objects.

## Approach

Invariant: composing a document never mutates baked metadata; two compositions from the same app produce identical documents. Deep-copy the baked schema and components at the point they enter a document (`writeSchemas`, alongside the existing `writeProperty` copy). Do not rely on customizers being idempotent. Check the composers for any other place that inserts baked objects by reference (response and exception schemas, security schemes) and fix the whole class.

[...]

PR #1659, opened 2026-09-23
fix: resolve the #1646-#1656 issue campaign cycle

samchon:
## Claim

This draft pull request owns the complete accepted cycle of the handover #1658: #1646, #1647, #1648, #1649, #1650, #1651, #1652, #1653, #1654, #1655, and #1656, implemented in one pull request.

Verification is pending. No implementation exists yet; this body is written at claim time, so it carries no closing keywords. Each issue is closed by the commit that resolves it.

## Planned order

| Order | Issue | What |
| --- | --- | --- |
| 1 | #1646 | Go module pins of typia match the npm dependency; recurrence guard |
| 2 | #1656 | `go test -count=1` in the three `test:go` scripts |
| 3 | #1655 | non-finite `@x-*` values no longer abort generation |
| 4 | #1654 | `SwaggerCustomizer` edits no longer leak across compositions |
| 5 | #1647 | generated SDK compiles on parameter name collisions |
| 6 | #1650 | one wire key, one OpenAPI parameter and one SDK input |
| 7 | #1652 | `@UploadedFile()` / `@UploadedFiles()` / `@RawBody()` |
| 8 | #1648 | plain `@Query()` / `@Headers()` / `@Param()` validation |
| 9 | #1649 | named examples as Example Objects; 2.0 generation |
| 10 | #1653 | `decompose: false` header and 2.0 parameters |
| 11 | #1651 | guides describe vanilla and encrypted route visibility |

Out of scope: releases (#1657) and every `version` field.


samchon (commented):
**Commit 63f182678: link the typia 15 Go source (#1646)**

- Moved the typia Go pin to `v0.0.0-20260922081121-78124b0523b4` (= `v15.0.0^{}`) in all four go.mod/go.sum files; the module source equals npm typia 15's `native` directory except LICENSE; `GOWORK=off go mod tidy` changes nothing further.
- Migrated to typia 15's `MetadataDependency_IListener` (`File` only; this host declares no `dependenciesComplete`, so `Callee`'s fallback and `Unbounded` change nothing).
- Recurrence guard `TestTypiaGoPinMatchesNpmTypia`: declared pins, resolved versions, and pinned source vs npm. Mutations: one lagging `sdk/native` pin → killed (resolution alone did **not** catch it; MVS selects core's newer pin); all four on typia 14 → killed by the source diff.
- `TestTransformTypiaPluginRewritesLlmEvaluationCall` fails on the typia 14 pin (call left untransformed) and passes on 15. `tags.Example<null>` → `"example": null` pinned in swagger-parameters.
- Behavior change from typia 15, verified deliberate: callee declarations of every examined call (decorators included) now appear in `dependencies`; the envelope test moved its no-entry twin to a call-free file.

Resolves #1646.

samchon (commented):
**Commit 6c8d784ee: `go test -count=1` (#1656)**

All three `test:go` scripts pass `-count=1`; the development skill states why. Locally `packages/core/test` now runs in ~22 s and `packages/sdk/test` in ~30 s on Windows.

Resolves #1656.

samchon (commented):
**Commit 5a7e0c39f: non-finite metadata numbers become null (#1655)**

- Reproduced beyond the issue's JSDoc case: numeric literal types (`1e999`, `-1e999 | 1`), enum members, and type tag arguments (`Minimum`, `Default`, `Example`, `Examples`, `JsonSchemaPlugin`) all aborted with `json: unsupported value`.
- The walk sits at the single marshal point (`nestiaSDKMetadataLiteralText`), not in `nestiaSDKJsonSchemaLiteral`: constants and type tags live in the metadata literal, which no schema normalizer sees.
- Numbers are matched by reflect kind, since the checker's are typescript-go's named `jsnum.Number` (first attempt with `case float64` missed them).
- Mutations: no walk / float64-only / no ordered-object handling → killed. Writing a Go nil instead of `LiteralFactory_Null` survives. That mutant is equivalent for every shape typia produces, because ordered objects hold schemas directly and never numbers. The marker stays because it prints null in every container, as `JSON.stringify` does.

Resolves #1655.

samchon (commented):
**Commit e1fbfd058: composed documents share no objects (#1654)**

- Copy where composition itself mutates its input: baked schemas/components in `writeSchemas`/`writeProperty` (the readonly-array emender edits them), `config.tags` in `initialize` (composition pushes route tags into it).
- Detach the whole document (in-place deep copy) before any customizer runs: covers decorator examples, `@ApiExtension` values, configured servers/security/info, and any future by-reference insertion.
- The copy exposed an implicit coupling: decomposition merged each property's un-emended components over the document's emended ones (`test_swagger_decomposed_query_schemas` caught `x-readonly-array` vanishing); the merge now only adds missing entries.
- CLI path covered too: the `configurations` array config composes CommonModule twice in one process.

Resolves #1654.

samchon (commented):
**Commit 0e569a388: lossless bigint / non-finite metadata values (#1655 follow-up, #1646 surface)**

**Early-warning pass on 63f182678 / 6c8d784ee / 5a7e0c39f, each candidate reproduced before acting**

| # | Candidate | Verdict | Action |
| --- | --- | --- | --- |
| C1 | `test_swagger_absent_schema_fields` walks the whole document for nulls; the new non-finite `x-level` nulls contradict it | confirmed | census now lists the eight non-finite nulls |
| C2 | `clone: true` crashed on a nulled constant (`text is not iterable`), and `Minimum<null>` would not compile | reproduced (scratch clone feature) | metadata literal carries non-finite constant/tag values by name (`"Infinity"`); clone printers decode by constant type / tag target; `1e999` is the literal type for ±Infinity; a NaN bound (JSDoc `@minimum NaN` reaches the SDK as `null`) is written in the `TagBase` form |
| C3 | `const: null` in schemas | by design: the issue's `JSON.stringify` rule for document content | kept |
| C4 | typia 15 marshals bigint constants as bare JSON numbers → clone printed `5` for `5n`, digits lost past 2^53 | reproduced | bigint constant/tag values carried as digit strings; printed as bigint literals; `LiteralFactory.write(bigint)` wrote a *string* literal, now a bigint literal |
| C5 | typia 15 `Extras.Shared` unset → per-file cache | confirmed (perf only) | next commit sets it once per program, as typia's hosts do |
| C6 | guard comment says LICENSE is absent from the module zip; it is the reverse | confirmed | fixed |
| C7 | consumers whose npm typia resolves past the Go pin get the #1646 skew | confirmed architecture gap (ttsc `composes` never builds the composed plugin's installed module) | filed separately |
| C8 | `key()` accepts a same-version replace | contrived | no change |

The document keeps #1655's `null`: only the internal metadata literal (constant and tag *values*) is lossless, because the SDK writes types back from it. Covered by `TestSyntheticNonFiniteValuesBecomeNull` (metadata side) and the new `clone-literal-values` feature (cloned DTO equals its source at compile time; tagged types give the same `typia.is` verdicts on boundary values).

Upstream typia defects found while reproducing (not nestia's; noted for the maintainer):
- a bigint type tag argument beyond ±2^53 is silently dropped (`bigint & tags.Maximum<9007199254740993n>` carries no tag);
- `ExpressionFactory.Number` formats non-finite numbers with Go's `fmt.Sprint`, so `typia.random<{ a: 1e999 }>()` emits `Inf` and throws `ReferenceError`.

samchon (commented):
**Commit b0ce2a6b5: SDK identifiers kept apart from user names (#1647)**

- One per-route naming pass (`SdkHttpParameterProgrammer.getNames`) shared by the SDK function, `path()`, `simulate()`, and the JSDoc `@param` tags. Fixed names are per scope and per config, so non-colliding routes are byte-identical (e.g. a route merely *named* `location` does not rename `path()`'s `location` local).
- Positional user parameter renamed only on a fixed-name collision; keyword `props` keys never renamed; SDK-own identifiers escape.
- Same class across WebSocket (+ existing bug: `path()` read `route.query.name` while declaring `query`) and MCP (tool names vs `client/args/raw/result/first`).
- Found in the touched functions: keyword-mode optional query object read as a bare identifier; `@setHeader value x-value` emitted `connection.headers.x-value` (now quoted, the migrate generator's precedent).
- Coverage: `sdk-name-collision`, `sdk-name-collision-keyword` (simulate+assert+json), `mcp-name-collision`. Local runs: both collision features, keyword/simulate/propagate/clone/websocket/mcp/headers/query groups all green.
- R2: the `mcp` feature listened on a hardcoded 37000 while the harness assigns ports (index 0 also gets 37000) → now `TEST_SDK_PORT`.

Resolves #1647.

samchon (commented):
**Commits 3dfaf735e..8b5a00bf1: early-warning follow-ups and #1648 #1649 #1650 #1651 #1652 #1653**

Early-warning pass on e1fbfd058 / 0e569a388 / b0ce2a6b5 (each reproduced or traced before acting):
- 18508cb16 (Refs #1654): customizers read their operation at call time, so a moved path crashed later ones; the JSON-copy detach threw on bigint examples. Operations are now resolved once after detach; the copy is structural (arrays and plain objects only, `__proto__`-safe, cycles kept).
- d78190453 (Refs #1655): tag values were decoded by `tag.target`, the tagged type, so `bigint & Sequence<1>` printed `Sequence<1n>`. Encoded values now carry `encoding`; Go test renamed to what it asserts.
- 7f5f99804 (Refs #1647): `path()`/WS locals escaped against names they cannot shadow (C1); WS JSDoc kept controller names (C2); route names shadowing module-scope bindings (`tags` import, `typia`, `String`, `exports`) broke the file (C3). Such a route is declared under a local name and `export { _x as x }`.
- 3dfaf735e (Refs #1646): typia's `Extras.Shared` set once per program (C5).

Issues:
- 49f0734b4: the error report threw on the first function and printed the accessor only when missing.
- 5994c786b #1652: `@UploadedFile()`/`@UploadedFiles()`/`@RawBody()` are diagnostics (multer field names unreadable; raw body untyped); `RAW_BODY` (12) was parsed as `1`.
- aad771e42 #1650: one wire key declared by a field and an object parameter is a diagnostic; headers compared case-insensitively.
- a2120eb99 #1648: typia's HTTP rules baked per parameter in Go, analyzed afresh (the shared collection had already escaped `Date` to string), picked by decorator. Dynamic-key query/header paths retired (#1648 lists them as error cases; v11 rejected them).
- 5ff483a3b #1651, d31615504 #1653, a4f70fa71 #1649 as titled.

Verification so far: `go test -count=1 ./...` green in packages/core/test and packages/sdk/test; full local test-sdk 163/166, the three failures being the clone NaN fixture and the collision JSDoc expectations, fixed in the next push along with a Go-side `@param` tag-shape regression they exposed (WebSocket `@param` lines were always dropped since v13). CI: format and migrate failed on this head (a comment wrap; the migrate test's legacy document literal claimed the Example Object type); both fixed in the next push.

samchon (commented):
**Commits baac09913..127558be0: follow-ups from the full local run and the early-warning pass on 3dfaf735e..8b5a00bf1**

- baac09913 (Refs #1647): the Go contributor wrote a `@param` tag as one text part since v13, while every reader expects TypeScript's `parameterName` / space / `text` parts, so WebSocket functions lost every `@param` line (and 7f5f99804's renaming never applied). The positional collision test's expectations are corrected (`_query` / `__query`). Go test `TestSyntheticParamTagParts`.
- 662c76791 (Refs #1647): module-scope aliases also avoid each other; WebSocket (`exports`) and MCP (`JSON`) aliasing now covered.
- 74533a30d (Refs #1648): a comment-tag violation named its property as `IType[key]`; now `ITagged.n`.
- 94da6d36c (Refs #1655): the NaN-bound clone fixture reached the printer only through the dropped-parameter flaw a2120eb99 closed; fixture and the printer's dead NaN handling removed.
- 465b15b9b (Refs #1649): #1649's invariant covers every nestia decorator, so Swagger 2.0 now also leaves out `x-nestia-encrypted`, form body descriptions and form object attributes, and derives form requiredness from its fields; each reproduced against typia's `SwaggerV2Downgrader` first.
- 2e1a963ff (Refs #1649): migrate reads `externalValue` / `serializedValue` / `dataValue` Example Objects; the named-examples test now matches the printer's layout and passed locally (`tests/test-migrate`, full run).
- 127558be0: development skill no longer pins test counts.

Local evidence at this head: `go test -count=1 ./...` green in packages/sdk/test and packages/core/test; test-sdk reruns green for name-collision (3), swagger-example, openapi_v2, query-decompose-false, multipart-form-data, the three error-plain features, clone (all 11), encrypted (2), plus the earlier full run (163/166, the three failures fixed here); test-migrate green.

Early-warning notes not acted on, with reasons: `@TypedQuery.Body()`'s SDK query rule uses `AllowOptional: true` where core uses `false` — moot, the core transform enforces the typed decorator first. Newly rejected vanilla types (dynamic-key `@Headers()` such as `IncomingHttpHeaders`, `any` field queries) now also fail `NestiaSwaggerComposer.document()` at run time — intended by #1648 and v11's behavior; recorded for the release notes.

Filed from the campaign's surface, outside this cycle's causes: #1662 (clone tag names), #1663 (consumer typia drift).

samchon (commented):
**Commits 8f6464a45..dcb38abc5: early-warning pass on baac09913..127558be0, and Self-Review round 1**

Self-Review round 1 (against 127558be0, whole base-to-head diff, 174 files): no functional defect beyond the early-warning findings; three documentation corrections applied in dcb38abc5 (stale 13-file / 155-feature counts in the skills, the `INames` "written exactly as before" claim, the HTTP-rules comment). CI was fully green on 127558be0, `sdk` lane included. Round 2 starts on this head.

Early-warning findings (each reproduced before acting):
- 8f6464a45 (Refs #1655): 94da6d36c's premise was incomplete — a NaN enum member passed to a tag (`tags.Minimum<Weird.A>`) reaches the clone printer and printed `value: NaN` in type position. Guard restored; clone-literal-values covers it.
- 5da836a38 (Refs #1647): the Go JSDoc reader cut every tag at its first line and misread `@param {T} name` / `@param [name]`; fixed to TypeScript's reading, Go test extended.
- 64751033c (Refs #1649): Swagger 2.0 still refused `File[]` / `File | null` form fields and non-JSON success bodies beside JSON exceptions; now representable.
- 62cad90a1 (Refs #1649): migrate `dataValue` / `serializedValue` covered.
- dcb38abc5 carries `Close #1658`.

Mutation record (disposable checkout @127558be0, Go test module): escape-mode rule analysis, param-tag parts, tag `encoding`, comment-tag accessor, field array rule, param union rule — all killed.
Website build per the documentation skill (`pnpm package:tgz`, `npm install --force && npm run build` in `website/`) passed @127558be0; website sources unchanged since.

Local evidence at dcb38abc5: `go test -count=1 ./...` green in packages/sdk/test; test-sdk green for clone-literal-values, swagger-* (8), name-collision (3), exception (4), websocket (5); test-migrate green. The integration gate restarts on this head.

samchon (commented):
**Commits c8a9599ab..4b81e67cd: early-warning pass on 8f6464a45..dcb38abc5**

Findings, each reproduced before acting:
- c8a9599ab (Refs #1647): with 5da836a38's multi-line tags, readers that split at spaces glued a tag's first word to the next line (`@tag`, `@security`, `@operationId`, `@throws` with a bare status, and `@setHeader`/`@assignHeaders`, which reach generated SDK code). Readers now split at any whitespace; the header tags read their first line only. New swagger-example route and multi-line collision header tags cover it.
- e9a198651 (Refs #1647): JSDoc lines keep their indentation past the `* ` margin, as TypeScript does; `@param [opt=[1,2]]` defaults with brackets.
- c92966b18 (Refs #1649): `Array<File | Blob>` / `Array<File | null>` still failed 2.0; any files-only field is one file, optional when it may carry none, keeping title/description/deprecation.
- 4b81e67cd (Refs #1649): the Swagger guide lists what a 2.0 document leaves out or narrows.

Not acted on: none. Self-Review round 2 (dcb38abc5) found only these; round 3 runs on this head. CI was green on dcb38abc5; the integration gate restarts here (install, format:check, `pnpm test`, `package:tgz`, clean packed consumer).

Local evidence at 4b81e67cd: `go test -count=1 ./...` in packages/sdk/test green; test-sdk green for swagger-example, name-collision (3), multipart-form-data, exception (4), swagger-parameters.

samchon (commented):
**Commits eaba396b3..519bfe8fc: Self-Review round 3**

Round 3 (against 4b81e67cd) was done without the early-warning reader, which is no longer used; every finding below was reproduced directly.

- eaba396b3 (Refs #1647): compared the Go JSDoc reader with TypeScript 5.9's `getJsDocTags` / `getDocumentationComment` over eleven comment shapes. e9a198651 kept too much indentation in three (wrapped `@param`, two-space margin, description on the opening `/**` line). The reader now takes TypeScript's margin; all eleven match except TypeScript's retained `\r` in CRLF comments. A new Go test pins five shapes to TypeScript's output.
- 519bfe8fc (Refs #1649): a `File[] & tags.MinItems<1>` form field stays a required 2.0 file (typia rejects the empty array); the 2.0 guide note moves under "OpenAPI versions".
- Also re-checked every JSDoc tag reader in the SDK for single-line or single-space assumptions: none remain.

Evidence at 519bfe8fc: Go test modules green (sdk, core); test-sdk green for swagger-example, name-collision (3), swagger-parameters, exception (4), websocket (5). CI green on 4b81e67cd. The integration gate restarts on this head; round 4 starts now.

samchon (commented):
**Commits 601e642b8..3b078a906: Self-Review rounds 4 to 6**

Each round covered the whole base-to-head diff: code, fixtures, tests, the website guides, and the agent skills. Every finding was reproduced before it was fixed.

- 601e642b8 (Refs #1647, round 4): a wrapped `@param` description of a renamed parameter (`_props`, `props.x`) was indented by the controller's parameter name, so its continuation lines fell out of line. Indentation now follows the name the SDK function declares; sdk-name-collision wraps one and asserts the alignment.
- 73d20e40f (Refs #1650, round 5): a function-level contradiction, which the #1650 duplicated-key check now also files, printed `Controller.method() from :`. An empty origin is left out in both reports; parameter-error-duplicated-key expects `DuplicatedController.fields():`.
- e17ad5732 (Refs #1647, round 6): in keyword mode, a WebSocket route named `query` or `provider` renamed its public `props` key to `_query` / `_provider` although a key shadows nothing. The keys now yield only to path parameters of the same name; positional parameters keep yielding to the route name, which they would shadow. Both collision features gain a `provider` route; reverting the fix fails the keyword feature's compile (`'provider' does not exist in type 'Props'`).
- 3b078a906 (Refs #1648, round 6): the TypedQuery guide said the diagnostic names the property; a union of objects has none, and the report names the type.

Checked and left as is (round 6):
- TypeScript 5.9 also splits a JSDoc tag at a line-start `@` inside a fenced code block, so the Go reader matches it there. TypeScript also splits at a mid-line ` @word`, which cuts the description off; the Go reader keeps the text, and the pre-v13 behavior is not worth reproducing.
- The WebSocket writer prints a multi-line tag without the continuation indent the HTTP writer adds. That is master's behavior, unchanged here, and TypeScript reads both the same way.

Evidence: sdk-name-collision and sdk-name-collision-keyword pass at e17ad5732 (the mutation check above was run on this pair). The integration gate restarts on 3b078a906, and round 7 starts now.

samchon (commented):
**Commits ff69f48bb..769c14fba: Self-Review round 7**

Round 7 checked the JSDoc path end to end against TypeScript 5.9. It covered the Go reader, both SDK writers, and TypeScript reading the generated SDK back.

- ff69f48bb (Refs #1647): TypeScript's margin for a `@param` continuation is the column of the description, past `{Type}`, the name or `[name=default]`, and any spaces. When the description starts on the next line, the margin is the tag's column. The Go reader used the name column, so a description continued in line with its first line (the aligned style formatters produce) kept 6 stray spaces in the parameter description. It also kept a `@returns {Type}` in the success response description, where TypeScript and pre-v13 nestia leave the type out. Six new margin-test cases come from TypeScript's own output for the same comments; disabling either rule fails them.
- 769c14fba (Refs #1647): multi-line tag text reaches the SDK writers again. The HTTP writer trimmed every continued line, which flattened `@example` code, and the WebSocket writer did not indent continued lines. `FilePrinter.jsDocTag` now writes the text so TypeScript reads it back exactly. TypeScript 5.9 reads the generated `read`, `provider` and `props` functions' tags as equal to their controllers'. A WebSocket route without a description no longer opens its comment with a blank line. swagger-example and both collision features pin the layout; disabling the writer's next-line branch and margin fails both.

Checked and left as is:
- TypeScript turns `@returns {@link X} …` into a separate `link` tag, and `@see name` followed by a next-line text keeps the ` * ` in its text. These are TypeScript bugs, and the reader does not reproduce them. `@see` and `@template` margins affect only the indentation of the re-emitted SDK comment; nestia reads neither tag.
- The exception writer keeps only the first line of a `@TypedException` description. That is master's behavior.

Gate note: gate8 (@3b078a906) failed `swagger-watch` ("initial swagger watch generation timed out", 90s). It overlapped my local feature runs and a Go module run. Alone at 769c14fba the feature passes in 64s, and CI's test job passed on 3b078a906. Gate9 now runs on 769c14fba with no concurrent local test-sdk work.

Evidence at 769c14fba: SDK Go module green (85s); sdk-name-collision, sdk-name-collision-keyword, swagger-example and swagger-watch pass. Round 8 found nothing further.

samchon (commented):
**Self-Review round 8 at 769c14fba: clean**

Round 8 re-read the whole base-to-head surface (code, fixtures, tests, guides, skills) against 769c14fba. It found no sound improvement or new issue. Checks beyond the diff reading:

- Core: typia 15's `MetadataDependency_IListener` falls back from a nil `Callee` to `File`, and ignoring `Unbounded` suits a host that only widens invalidation, which nestia's envelope does (the consumer unions it with the compiler graph). The shared typia cache is built once per program in both entry points and is concurrency-safe.
- SDK analysis: `@ignore` removes a method before any parameter analysis, so the vanilla-payload diagnostics' advice holds. Report aggregation, the document detach and clone, the copied bakes, and the bigint, NaN, and Infinity literal writers are coherent with their callers.
- JSDoc: every reader of a `@param` tag takes the `parameterName` and `text` parts. A description starting on the next line reaches the SDK trimmed through `doc.Params`, and the WebSocket writer reprints the raw tag text faithfully. Typed `@returns` text feeds the success response description without the type. The e2e generator uses only user parameter names as `props` keys, so the naming rules never reach it.
- Guides: no page describes the WebSocket keys or the SDK comment layout; the Swagger guides use untyped `@returns`.

Pending before merge: CI and the integration gate (gate9) at 769c14fba.

samchon (commented):
**Self-Review round 9 at 769c14fba: acceptance traceability, clean**

Every acceptance item of #1646–#1656 was checked against the head.

| Issue | Acceptance | Evidence |
| --- | --- | --- |
| #1646 | plugin links the typia 15 pin; `llm.evaluation` transforms; `Example<null>` documents `null`; guard in CI | `TestTypiaGoPinMatchesNpmTypia` (four modules, source diff against npm), `TestTransformTypiaPluginRewritesLlmEvaluationCall`, swagger-parameters `single.example` |
| #1647 | collision features' SDK, simulator, and e2e compile and call every route; no rename without a collision | sdk-name-collision (+keyword), mcp-name-collision; scope-local naming pinned by `test_sdk_name_collision_path_locals` |
| #1648 | the four routes fail with diagnostics; valid vanilla shapes generate; error features for query, headers, param | query/headers/param-error-plain; swagger-parameters `vanilla/:kind` and the vanilla `@Query()` DTO with a nullable property |
| #1649 | named examples as `{ value }`; 2.0 generates; migrate round trip | test_swagger_example, test_swagger_example_v2, test_migrate_nest_named_examples |
| #1650 | diagnostic naming `keyword` and `X-Tenant`, case-insensitive headers, error feature | parameter-error-duplicated-key |
| #1651 | no guide claims vanilla or encrypted routes are absent; website build; typos | CI `website` and `spell-check` green |
| #1652 | the three decorators fail with a diagnostic naming the alternative; `@Req()` routes unchanged | payload-error-vanilla; the body and beautify features keep `@Req()` beside typed parameters |
| #1653 | 3.x headers split, query `style: form` / `explode: true`; 2.0 without `$ref` or object types; upgrade keeps every parameter | test_swagger_undecomposed_parameters |
| #1654 | three compositions equal, each edit once; a caller's edit does not leak | test_swagger_composition_isolation, test_swagger_customizer_per_document |
| #1655 | the interface generates with non-finite values as `null`; a Go test for `NaN`, `Infinity`, `-Infinity`, `inf` | `TestSyntheticMetadataValuesJsonCannotHold`, test_swagger_non_finite_extensions |
| #1656 | `test:go` under a minute on Windows; `-count=1` everywhere | gate9: sdk module 48.4s, core 36.6s; all three scripts pass `-count=1` |

Deliberate refinements, recorded for readers of the issues:
- #1655: JSON documents write `null` as `JSON.stringify` does. The metadata literal carries a constant's or tag's non-finite value by name and a bigint as digits, because the cloned DTO is typed from it and `null` lost the sign.
- #1647 asked for byte-identical output where nothing collides. The naming change keeps that. The round-7 writer fix (769c14fba) changes the comment layout only where it was wrong: multi-line tag text with its own indentation, and a description-less WebSocket comment that opened with a blank line.

samchon (commented):
**Decisions and deliberate-behavior record (handover #1658)**

The handover left four decisions open and named three earlier choices whose reasoning had to be checked before behavior changed. The whole record is here, so a reader need not assemble it from the commit ledger.

### Open decisions

| Issue | Decision | Evidence and precedent |
| --- | --- | --- |
| #1648 | Take the verdict from typia's own Go validators. The SDK contributor bakes, per route parameter, the violations of each HTTP category (query, headers, param, field, formData), and `ReflectHttpOperationParameterAnalyzer` picks the category its decorator selects. No TypeScript port. | typia exports `HttpQueryProgrammer.Validate`, `HttpHeadersProgrammer.Validate`, and `HttpFormDataProgrammer.Validate`, so these rules cannot drift from what the core transform enforces on `@Typed*`. Only the path rule (unexported in typia) and the field-named rule (v11's own) are written in nestia, each pinned with accepted twins in `TestSyntheticParameterHttpRules`. The type is analyzed afresh, as typia's HTTP programmers do: the SDK's shared collection had already escaped `Date` to a string, which would have let a `Date` query pass. The two decomposition paths the issue flagged are settled. A field-named parameter is one key and is never decomposed, and the dynamic-key query/header paths are retired, because those types are now diagnostics, as in v11. |
| #1649 | Named examples become OpenAPI Example Objects (`{ value }`) at analysis (`SwaggerExampleAnalyzer`). Swagger 2.0 leaves out what it cannot hold. `@nestia/migrate` reads Example Objects, with a raw-value fallback. | OpenAPI 3.x `examples` maps hold Example Objects. Swagger UI and typia's downgrader read `value`, so the raw shape documented no example. Each 2.0 omission was reproduced against typia's `SwaggerV2Downgrader` before it was made. |
| #1650 | Option (a): a diagnostic when a field parameter and the object parameter of its category declare one wire key, with headers compared case-insensitively. | NestJS hands both parameters the same `req.query.keyword` and lowercases header names, so the two declarations cannot differ on the wire. Option (b) would silently merge two declarations that may disagree in type. Precedent: two field parameters of one header name were already a contradiction (`Duplicated field names of headers are not allowed.`). |
| #1652 | Option (a): `@UploadedFile()`, `@UploadedFiles()`, and `@RawBody()` are diagnostics naming the supported alternative, or `@ignore`. | The multer field names live in the interceptor's closure (`FileInterceptor("file")`), which no metadata exposes, so option (b) cannot produce an SDK that sends the files. A raw body has no type to describe. `@ignore` removes a method before any parameter analysis, so the advice holds. |

### Earlier choices, and why their reasoning does not cover these cases

- **#1470's validator stubs.** #1470 short-circuited the `Http*Programmer.validate` slots because "the typia native transform already enforces their structural invariants". That holds for `@TypedQuery()`, `@TypedHeaders()`, `@TypedParam()`, and `@TypedFormData.Body()`, which the core transform runs typia's HTTP programmers on. No transform touches a vanilla `@Query()`, `@Headers()`, or `@Param()`, so for those nothing validated at all. The baked rules apply to every parameter; for a typed decorator they repeat the core transform's verdict with the same typia code. The stubs and their stale doc comments (`HttpQueryValidator.ts`, `HttpHeadersValidator.ts`) are removed.
- **Raw examples round-tripping through migrate.** The #1639 campaign kept raw values so a nestia document and `@nestia/migrate` stayed symmetric. That symmetry is kept, not dropped. Migrate now reads the Example Object nestia writes, and still reads the raw values of documents from earlier nestia versions, so a round trip reproduces the same `SwaggerExample` calls (`test_migrate_nest_named_examples`, run against the fixture document nestia generates). The raw shape itself violated OpenAPI, which the round trip never required.
- **The SDK escaping its own identifiers.** This is the convention #1647 follows, not one it overrides. `output`, `variables`, and the WebSocket `connection` were already escaped against user names. Now every SDK-own identifier is escaped per scope, only against the names that scope can see. A user's positional parameter is renamed only when its name is one the SDK cannot change (the route's public function and namespace, imports, globals). Positional callers never see that rename, and keyword-mode `props` keys are never renamed. Controllers without a collision generate the same names as before.

### Behavior changes for the release notes (#1657)

- `NestiaSwaggerComposer.document()` at runtime now throws for vanilla parameters of types the wire cannot carry, as `nestia sdk` / `nestia swagger` report them (#1648, v11's behavior).
- Named examples change shape from raw values to Example Objects (#1649).
- With `decompose: false`, a header object is decomposed anyway, and a 3.x query object gains `style: form` / `explode: true` (#1653).

samchon (commented):
**Merge gate at 769c14fba**

- **CI:** 21/21 checks pass at 769c14fba.
- **Self-Review:** round 7 found and fixed the JSDoc reader and writer issues. Rounds 8 (whole surface), 9 (acceptance traceability), and 10 (the generated SDK read as a consumer would) found nothing further.
- **Integration gate** (disposable checkout; `origin/master` 9a4d3d672 is an ancestor, so the prospective merge equals the head; re-read before merge, unchanged):
  - `pnpm install --frozen-lockfile` ✓
  - `pnpm format:check` ✓
  - `pnpm test` ✓: build, both Go test modules (sdk 48.4s, core 36.6s), and every `tests/test-*` workspace, including all test-sdk features; swagger-watch took 101s here, with no concurrent local load
  - `pnpm package:tgz` ✓
  - Clean packed consumer (tarballs from this head, npm `typia@^15`, `ttsc` 0.28.6, `typescript` 7.0.2): `nestia swagger` ✓ (3.x named examples as `{ value }`, 2.0 document generated), `nestia sdk` ✓.
  - The consumer's `ttsc` build failed first because of my fixture. It listed typia's, nestia core's, and nestia sdk's transforms as three enabled plugins, which ttsc refuses to compose into one emit pass, and which the setup guide never asks for (it says to list none, or to disable typia's when overriding options). The fixture also lacked the `rootDir` TypeScript 7 requires with `outDir`. Corrected, the same tarballs build on both documented forms: no `plugins`, and the guide's override (`typia` disabled, `@nestia/core/native/transform.cjs` with `validatePrune`). The emitted controller carries typia's validators, and pruning under the override. Neither package's ttsc plugin manifest changed in this pull request.
- **Mutation records:** each round-7 fix and the earlier recorded fixes fail their tests when reverted.
- **Closing lines:** `git log origin/master..HEAD` carries `Close` lines for #1646–#1656 and #1658, each with a surviving fix at the head.

After merge, the same gate runs on the merge commit.

samchon (commented):
**Post-merge verification at 78ddae1ac**

The squash commit's tree equals the verified head's (`769c14fba^{tree}`), and the same gate ran on it:

- `pnpm install --frozen-lockfile` ✓
- `pnpm format:check` ✓
- `pnpm test` ✓ (every workspace; swagger-watch 78s)
- `pnpm package:tgz` ✓
- Clean packed consumer: `nestia swagger` ✓ (3.x `{ value }` named examples, 2.0 document), `nestia sdk` ✓, `ttsc` build with no `plugins` ✓, and with the setup guide's override ✓.

Correction to the merge-gate review: it said the emitted controller showed pruning under the override. The consumer has no `@TypedBody()`, so `validatePrune` has nothing to act on there. What the override demonstrably changes is `stringify: "validate.log"`. The default build serializes responses through `assert` (`_ao0`, `_assertGuard`), and the override build through `validate` (`_vo0`, `_validate`, `_validateReport`).

Issues #1646–#1656 and #1658 are closed as completed. #1657 (release), #1662, and #1663 remain open.

